PR #17, opened 2020-07-22
State selection page UI

casprice:
image


src/main/webapp/list.html
```diff
@@ -0,0 +1,136 @@
...
+          <option value="state:fl">Florida</option>
+          <option value="state:ga">Georgia</option>
+          <option value="state:hi">Hawaii</option>
+          <option value="state:ia">Iowa</option>
+          <option value="state:id">Idaho</option>
+          <option value="state:il">Illinois</option>
+          <option value="state:in">Indiana</option>
+          <option value="state:ks">Kansas</option>
+          <option value="state:ky">Kentucky</option>
+          <option value="state:la">Louisiana":
```
gianelgado12, line 52: Small typo in the Louisiana option

src/main/webapp/list.html
```diff
@@ -0,0 +1,136 @@
...
+    </a>
+    
+    <div class="container" id="content">
+      <div class="row page-title-text" id="dropdown-title">
+        <p class="col-lg-8 col-lg-offset-2 col-md-8 col-md-offset-2 col-sm-8 col-sm-offset-2">
+          <b>To explore the upcoming elections you can vote in, select the state you live in:</b>
+        </p> 
+      </div>
+      <div class="row" id="state-dropdown">
+        <select id="select-state" name="states" onchange="listElections()">
```
gianelgado12, line 32: Is there anyway to move arrow in the selection box further away from the edge of the box?

casprice: I didn't spend a long time looking into alternatives, but a few Google searches told me that dropdown menus are pretty hard to style and stuff like height of the dropdown and arrow are dependent on the browser, not CSS :/

rogerlaucorp:
Can you send PRs to me and julian too?

casprice:
Already approved but I made some more changes (hard-coded list of elections)
image

src/main/webapp/list.html
```diff
@@ -0,0 +1,177 @@
...
+            <p>August 4, 2020</p>
+          </div>
+          <div class="learn-more-button">Learn more</div>
+        </li>
+        <li class="list-group-item" id="election-item">
+          <div class="election-quick-details">
+            <p>Other Election Title</p>
+            <p>August 40, 2020</p>
+          </div>
+          <button class="learn-more-button">Learn more</button>
```
gianelgado12, line 121: Is this also supposed to be a div instead of a button?

src/main/webapp/script.js
```diff
@@ -0,0 +1,7 @@
+function listElections() {
```
yang-julian, line 1: Maybe omit this file until the script does something meaningful?

src/main/webapp/list.html
```diff
@@ -0,0 +1,177 @@
+<!DOCTYPE html>
```
yang-julian, line 1: (couldn't figure out how to make a file comment)

Can we rename this file to be a bit more descriptive?   maybe at the very least "list_state.html" or "state_selection.html" to narrow down what the purpose of this webpage is :)

src/main/webapp/list.html
casprice: Whoops, good catch

src/main/webapp/electionlist.html
```diff
@@ -0,0 +1,177 @@
...
+            <p>Other Election Title</p>
+            <p><em>August 40, 2020</em></p>
+          </div>
+          <div class="learn-more-button">Learn more</div>
+        </li>
+      </ul>
+    </div>
+
+    <div class="container-fluid" id="content-padding">
+      &nbsp;
```
yang-julian, line 127: I have no idea the answer to this one... do we need to keep this space?

src/main/webapp/electionlist.html
```diff
@@ -0,0 +1,177 @@
...
+    </a>
+
+    <div class="container" id="content">
+      <div class="row page-title-text" id="dropdown-title">
+        <p class="col-lg-8 col-lg-offset-2 col-md-8 col-md-offset-2 col-sm-8 col-sm-offset-2">
+          <b>To explore the upcoming elections you can vote in, select the state you live in:</b>
+        </p> 
+      </div>
+      <div class="row" id="state-dropdown">
+        <select id="select-state" name="states" onchange="listElections()">
```
yang-julian, line 32: Do we have this method already? If not maybe omit the onchange until we add the corresponding js function.

src/main/webapp/electionlist.html
casprice: Yeah, as of now the contents of the page get blocked blue bar on the bottom of the screen without the space. We're planning to later make the blue bar a footer so this isn't a problem.

gianelgado12: I added that space so if the user changes their viewport to something that's more tall than wide (for example if they snap the window to one half of the screen) then they can still scroll down to see the content but I think we're planning on making the about section a footer instead of a fixed element so it'll probably become unnecessary if we go down that path.

PR #26, opened 2020-07-23
Election entity modeled off of the description in the spec.

anooshrees:


src/main/java/com/google/sps/data/Election.java
```diff
@@ -0,0 +1,43 @@
...
+package com.google.sps.data;
+
+import java.util.HashSet;
+
+/** An item on a todo list. */
+public final class Election {
+
+  private final String name;
+  private final String scope;
+  private final HashSet<long> positions;
```
rogerlaucorp, line 24: Does this work? I thought it need to be HashSet<Long> (autoboxed).

rogerlaucorp: Same below.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -0,0 +1,43 @@
...
+    this.date = date;
+    this.positions = new HashSet<long> ();
+    this.propositions = new HashSet<long> ();
+  }
+
+  public void addPosition(int positionID) {
+      this.positions.add(positionID);
+  }
+
+  public void addProposition(int propID) {
```
rogerlaucorp, line 40: propositionId instead of propID. 

Java style guide enforces camelCase, and we generally discourage abbreviations.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public Election(String name, String scope, String date) {
+    this.name = name;
+    this.scope = scope;
+    this.date = date;
+    this.positions = new HashSet<long> ();
+    this.propositions = new HashSet<long> ();
+  }
+
+  public void addPosition(int positionID) {
```
rogerlaucorp, line 36: positionId instead of positionID per style guide.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.HashSet;
+
+/** An item on a todo list. */
```
rogerlaucorp: I don't think this is an item on a TODO list.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -0,0 +1,43 @@
...
+public final class Election {
+
+  private final String name;
+  private final String scope;
+  private final HashSet<long> positions;
+  private final String date;
+  private final HashSet<long> propositions;
+
+  public Election(String name, String scope, String date) {
```
rogerlaucorp, line 28: I may have mentioned this before: I'm okay with having a simple data container class like this given the time constaints of the project. But for a professional project, we would usually enforce that you use the Builder pattern:
https://refactoring.guru/design-patterns/builder/java/example

Usually autogenerated to reduce boilerplate using AutoValue:
https://github.com/google/auto/blob/master/value/userguide/index.md

You can just ACK (acknowledge) this comment.

src/main/java/com/google/sps/data/Election.java
rogerlaucorp, line 40: So you can never clear the list? That's fine, just double checking that's the intention.

anooshrees: I just delete the entire election entity; I don't anticipate needing to clear but can always add that functionality later.

src/main/java/com/google/sps/data/Election.java
anooshrees: ACK

src/main/java/com/google/sps/data/Election.java
anooshrees: Resolved in later commit.

src/main/java/com/google/sps/data/Election.java
anooshrees: You are correct -- happy to add that functionality later should it be necessary.

src/main/java/com/google/sps/data/Election.java
anooshrees: Resolved in later commit.

PR #27, opened 2020-07-23
Elections datastore

anooshrees:


anooshrees:
> FYI (you can just ACK this) I think giving you data-classes like this would have been the main benefit of using the Google API Client Library (https://developers.google.com/api-client-library/java)I linked the other day, instead of having to manually create and populate them. Like Google's official version of this class would be:
> 
> https://developers.google.com/resources/api-libraries/documentation/civicinfo/v2/java/latest/com/google/api/services/civicinfo/model/Election.html
> 
> Notably, sometimes there's reasons to want to your own domain objects anyway. See:
> https://www.geeksforgeeks.org/adapter-pattern/
> as one example. Another example would be if the underlying API is unstable, and you don't want API changes to break your code, you'd have an insolation layer that you control which is stable.

One primary reason I chose not to do this is because, in the design spec, we established that Election entities would contain HashSets of the ids of candidates and propositions on their ballots, which would require substantial changes to the class you provided above.

anooshrees:
FYI, here are the details for the cron job that will call doPut():
image

src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -98,13 +119,48 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+        JSONObject obj = new JSONObject(elections);
+        JSONArray electionData = obj.getJSONArray("elections");
+        int numElections = electionData.length();
+
+        for (int i = 0; i < numElections; ++i) {
+            JSONObject currentElection = electionData.getJSONObject(i);
+            Entity electionEntity = new Entity("Election");
+
+            electionEntity.setProperty("id", currentElection.getLong("id"));
```
yang-julian, line 133: Hm. Is there a reason why we didn't use this value as the key/id for the Entity?

src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -49,11 +62,20 @@
   private static final String baseURL = "https://www.googleapis.com/civicinfo/v2/elections?key=";
   private static final Logger logger = Logger.getLogger(ElectionServlet.class.getName());
 
-  // TODO(anooshree): Change GET to PUT and store retrieved information in the database
   @Override
-  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  public void doPut(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        // Deleting the queries from yesterday in the case that they are outdated
```
yang-julian, line 68: Could we update them (e.g. by Election#id) instead of deleting them?

src/main/java/com/google/sps/servlets/ElectionServlet.java
anooshrees: I was thinking of the "id" value of the election entity as the ID provided by the API and a feature of the election (similar to its name), while the ID provided by datastore is generated when the election is added to Datastore. 

Do you think we should change it to the key/id for the Entity given that information?

src/main/java/com/google/sps/servlets/ElectionServlet.java
anooshrees: Sure -- I did this to handle the case where elections that were present the previous day are no longer available. Because Datastore access entities by ID and not by index, we wouldn't know the difference in the number of elections between days and would thus have to compare the entries side by side, so I figured deleting and repopulating the entities would be faster.

src/main/java/com/google/sps/servlets/ElectionServlet.java
yang-julian: I think it makes sense to use the ID provided by the API as the ID in our datastore.  There's no need to have datastore generate a new one if we can re-use the ID we already have.  Though granted, I haven't checked if the ID from Civics api conforms with the datastore key requirements, so that might be a blocker.

anooshrees: I'll check right now!

src/main/java/com/google/sps/servlets/ElectionServlet.java
yang-julian: I see.  Yeah I think that approach is probably ok for now.  We can always change it later if there are issues.

Re elections that have passed, I think the frontend/ client can always just filter out elections by election day as well.

src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -98,13 +119,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+        JSONObject obj = new JSONObject(elections);
+        JSONArray electionData = obj.getJSONArray("elections");
+        int numElections = electionData.length();
+
+        for (int i = 0; i < numElections; ++i) {
+            JSONObject currentElection = electionData.getJSONObject(i);
+            Entity electionEntity = new Entity("Election");
+
+            /* The "id" of an Election Entity is stored as a property instead of
```
yang-julian, line 133: nit: Can you follow one of these conventions for block comments?

https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -98,13 +119,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+               replacing the Datastore-generated ID because  Datastore may
```
yang-julian, line 134: please remove extra space.

anooshrees: Are you referring to the carriage return?

src/main/java/com/google/sps/servlets/ElectionServlet.java
anooshrees: resolved in PR for "Comment Formatting".

PR #28, opened 2020-07-23
List elections from the data returned from ElectionServlet

casprice:


src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -14,29 +14,19 @@
 
 package com.google.sps.servlets;
```
casprice, line 16: Quick note about this file: I didn't change any code, the file just "rewrote" because the Java formatter fixed the indentation

src/main/webapp/script.js
```diff
@@ -0,0 +1,162 @@
...
+      electionList.forEach((election) => {
+        let ocdId = election.ocdDivisionId;
+        let stateId = ocdId.indexOf(OCD_STATE_TITLE);
+        let districtId = ocdId.indexOf(OCD_DISTRICT_TITLE);
+
+        // ocdDivisionId is in the form "ocd-division/country:us/state:<state_code>"
+        // For example, Lousiana's ID is "ocd-division/country:us/state:la"
+        if (stateId !== -1) {
+          // If the state text is found in ocdDivisionId, then this is a state election.
+          if (ocdId.substring(stateId, stateId + OCD_STATE_TITLE.length + STATE_CODE_LENGTH)
```
yang-julian, line 43: No need to change this, but normally we would use regular expressions (regex) for this, especially since the format is fairly straight forward.
The pros of regex is you can actually extract the string you are looking for directly, if your search expression matches.   Unfortunately for more complicated patterns, regex can get pretty tricky and difficult to get right.

Some more info here: https://www.regular-expressions.info/

Happy to chat more if you're interested in using regex.

src/main/webapp/script.js
```diff
@@ -0,0 +1,162 @@
...
+      // to later display the details to the user.
+      electionList.forEach((election) => {
+        let ocdId = election.ocdDivisionId;
+        let stateId = ocdId.indexOf(OCD_STATE_TITLE);
+        let districtId = ocdId.indexOf(OCD_DISTRICT_TITLE);
+
+        // ocdDivisionId is in the form "ocd-division/country:us/state:<state_code>"
+        // For example, Lousiana's ID is "ocd-division/country:us/state:la"
+        if (stateId !== -1) {
+          // If the state text is found in ocdDivisionId, then this is a state election.
```
yang-julian, line 42: IMO this comment is not needed :)

src/main/webapp/script.js
```diff
@@ -0,0 +1,162 @@
...
+ */
+function buildElectionListElements(containerElement, elections) {
+  const electionListElement = document.createElement('ul');
+  electionListElement.className = 'list-group list-group-flush';
+  electionListElement.id = 'election-list';
+
+  elections.forEach((election) => {
+    const electionItemElement = document.createElement('li');
+    electionItemElement.className = 'list-group-item';
+    electionItemElement.id = 'election-item';
```
yang-julian, line 97: I thought we're not allowed to have multiple elements with the same html id?

src/main/webapp/script.js
```diff
@@ -0,0 +1,162 @@
...
+const COUNTRY_NAME = ' the United States';
+const NO_ELECTIONS_TEXT = 'No elections coming up soon.';
+const LEARN_MORE_BUTTON_TEXT = 'Learn more';
+const INVALID_DATE_TEXT = 'Invalid date';
+
+/**
+ * Fetch the list of upcoming elections and display relevant ones according
+ * to the state selected by the user.
+ */
+function listElections() {
```
yang-julian, line 16: Can you add a todo to write tests?

https://medium.com/@rishabhsrao/mocking-and-testing-fetch-with-jest-c4d670e2e167
This may help for how to mock the fetch() calls.

src/main/webapp/script.js
```diff
@@ -0,0 +1,162 @@
...
+      }
+  });
+}
+
+/**
+ * Build the HTML elements that make up the list of upcoming elections.
+ * @param {HTMLDivElement} containerElement div container for election list
+ * @param {JSON} elections JSON object containing the list of relevant elections.
+ */
+function buildElectionListElements(containerElement, elections) {
```
yang-julian, line 89: I feel like this section would be much more readable if we could use an HTML templating engine.  I'm looking more into that right now.
[Handlebars.js](https://handlebarsjs.com/guide/) is looking pretty promising.

src/main/webapp/script.js
```diff
@@ -0,0 +1,162 @@
...
+  fetch('/election')
+    .then(response => response.json())
+    .then((textResponse) => {
+      let electionList = JSON.parse(textResponse).elections;
+
+      // For every election returned, store it if the state matches or if national election,
+      // to later display the details to the user.
+      electionList.forEach((election) => {
+        let ocdId = election.ocdDivisionId;
+        let stateId = ocdId.indexOf(OCD_STATE_TITLE);
```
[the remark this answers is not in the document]
casprice: Yeah, will do

src/main/webapp/script.js
casprice: Oops I should've made that a class
Not sure why it still works hahah but I'll change it!

src/main/webapp/script.js
casprice: Sure, I'm planning to do tests tomorrow. Thanks for the link!

src/main/webapp/script.js
```diff
@@ -0,0 +1,33 @@
...
+      
+      buildElectionListElements(stateElections);
+  });
+}
+
+function buildElectionListElements(elections) {
+  const electionListContainerElement = document.getElementById('election-list-content');
+
+  const electionsInStateHeadingElement = document.createElement('h3');
+  electionsInStateHeadingElement.innerHTML = 'Elections coming up in Washington';
```
rogerlaucorp: I assume "Washington" is going to get replaced later? Can you either have a comment indicating where it'll get a replaced, or a TODO to dynamically populate it?

src/main/webapp/script.js
```diff
@@ -0,0 +1,33 @@
...
+          // TODO: update ocdDivisionId to cater to state vs. national elections
+          stateElections.append(election);
+        }
+      });
+      
+      buildElectionListElements(stateElections);
+  });
+}
+
+function buildElectionListElements(elections) {
```
rogerlaucorp: Does the 'elections' parameter ever get used in this function?

rogerlaucorp: Again leave a TODO if it's going to get used later but you just have this here right now.

src/main/webapp/script.js
```diff
@@ -12,22 +23,122 @@ function listElections() {
...
+    electionListElement.appendChild(electionItemElement);
   });
+
+  containerElement.appendChild(electionListElement);
 }
 
-function buildElectionListElements(elections) {
-  const electionListContainerElement = document.getElementById('election-list-content');
+/**
+ * Build the text that
```
rogerlaucorp: That what?

src/main/webapp/script.js
casprice: Sorry, old commit was using hard-coded information to check that the elements were showing up as expected. The Washington is gone now!

src/main/webapp/script.js
casprice: I just switched this implementation to Handlebars, so I won't be using this function anymore!

src/main/webapp/script.js
casprice: Haha, again old commit where I hadn't finished documenting. Should I be consistently sticking documentation & implementation in the same commit?

yang-julian: hm.. I think in terms of "git" commits, it probably doesn't matter. But the documentation & implementation should happen in the same pull request if that makes sense.

rogerlaucorp: Yeah that's fine. Sorry, github just makes it hard to review since I need to go through all the commit, but some of the commits overwrite each other.

src/main/webapp/script.js
```diff
@@ -39,124 +34,53 @@ function listElections() {
...
-}
-
 /**
  * Change the format of the date returned by the Google Civic Information
  * API to be the standard date format (ex. 2020-08-16 ==> August 16, 2020).
  * @param {String} apiDate date in the form YYYY-MM-DD
  * @return {String} the formatted date in the form Month DD, YYYY
  */
-function formatDate(apiDate) {
+Handlebars.registerHelper("formatDate", function(apiDate) {
```
yang-julian, line 70: nice!

src/main/webapp/script.js
casprice: I usually look it over through "Files changed", I think that has the most up-to-date code

src/main/webapp/script.js
```diff
@@ -22,8 +22,8 @@ function listElections() {
       let electionList = JSON.parse(textResponse).elections;
       // ocdDivisionId is in the form "ocd-division/country:us/state:<state_code>"
       // For example, Lousiana's ID is "ocd-division/country:us/state:la"
-      let stateRegex = /ocd\-division\/country\:us\/state\:[a-z][a-z]/;
-      let districtRegex = /ocd\-division\/country\:us\/district\:[a-z][a-z]/;
+      let stateRegex = /ocd\-division\/country\:us\/state\:([a-z][a-z])/;
```
yang-julian, line 27: \* Just FYI, no action required \*

A slight optimization is using [regex alternation](https://www.regular-expressions.info/alternation.html)
You can take advantage of the fact that the two regex is almost identical except for the fact that one uses "state" and the other uses "district".  So if we use this regex:
`/ocd\-division\/country\:us\/(state|district)\:([a-z][a-z])/`
** Note that I added an extra capture group surrounding (state|district) so that the alternate as a whole is treated as one entity.  You can try experimenting what happens when you don't have the parentheses.

It will match for both "ocd-division/country:us/state:la". and "ocd-division/country:us/district:dc"

For example: https://regex101.com/r/KlNoFT/1/tests [Regex101 is a great tool for testing your your regex, and for explaining what's going on in the regex]

casprice: Ohh thank you!

src/main/webapp/script.js
```diff
@@ -34,8 +33,8 @@ function listElections() {
 
         // Match the OCD Division ID with the regular expressions to check if this election
         // is a state election, district election, or national election.
-        let stateMatch = stateRegex.exec(ocdId);
-        let districtMatch = districtRegex.exec(ocdId);
+        let stateMatch = regex.exec(ocdId);
+        let districtMatch = regex.exec(ocdId);
```
yang-julian, line 37: I think you can just combine these into one variable right? since they have the same exact result :)

PR #30, opened 2020-07-24
Create barebones User class in order to store user data on Datastore.

anooshrees:


src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,53 @@
...
+  public void setState(String state) {
+    this.state = state;
+  }
+
+  public String getState() {
+    return this.state;
+  }
+
+  public void addElection(Election election) {
+    this.elections.add(election);
```
yang-julian, line 47: The "this" here is already implicit, so we can remove it.  same below

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.HashSet;
+
+/** A user who has provided us with location information as they navigate the site */
+public class User {
```
yang-julian, line 20: What is this class used for again? Can't we just store the user data in the javascript code?

PR #31, opened 2020-07-27
Cron job adjustments

anooshrees:
Add additional servlet that cron job calls, and repurpose /election servlet to only retrieve information from Datastore after /getElection stores it. Add necessary dependencies to the pom.xml file, and use Secrets API to hide API key during site deployment.


anooshrees:
This is the Secrets Manager API I refer to: https://cloud.google.com/secret-manager/docs/creating-and-accessing-secrets#secretmanager-access-secret-version-java

src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -14,160 +14,51 @@
...
-        StringBuilder strBuf = new StringBuilder();  
-        HttpURLConnection conn = null;
-        BufferedReader reader = null;
-
-        try {  
-            // Requires storing API key in environment variable.
-            URL url = new URL(baseURL + System.getenv("GOOGLE_API_KEY"));
-            conn = (HttpURLConnection) url.openConnection();  
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Query query = new Query("Election");
```
rogerlaucorp, line 41: Is there a "builder" version of query? If so, please prefer that. Builders are nice because they document variables and help prevent bugs related to variable mis-ordering.

src/main/java/com/google/sps/servlets/GetElectionServlet.java
```diff
@@ -0,0 +1,169 @@
...
+ * This servlet is used to update the gVote database with information on the ongoing elections that
+ * an eligible voter can participate in on a given day.
+ */
+public class GetElectionServlet extends HttpServlet {
+
+  private static final String baseURL = "https://www.googleapis.com/civicinfo/v2/elections?key=";
+  private static final Logger logger = Logger.getLogger(ElectionServlet.class.getName());
+
+  // This method is used to access the api key stored in gcloud secret manager.
+  public String accessSecretVersion(String projectId, String secretId, String versionId)
```
rogerlaucorp, line 55: Rename method to a higher level of abstraction. I.e. yes this function is accessing the secret version, but what does the secret version represent?

GetApiKey() might be more apt. Clients of this function can read the javadoc/implementation if they actually care to know how getting the API key is actually impelmented.

src/main/java/com/google/sps/servlets/GetElectionServlet.java
```diff
@@ -0,0 +1,169 @@
...
+      throws IOException {
+    try (SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
+      SecretVersionName secretVersionName = SecretVersionName.of(projectId, secretId, versionId);
+
+      // Access the secret version.
```
rogerlaucorp, line 60: Remove this comment. This is an example of a "bad" comment which describes the "what I'm doing" not the "why I'm doing it". I can just read the line below and figure out from "client.accessSecretVersion()" what the comment is saying. 

You could replace this comment explaining why you don't want to reveal the API key and instead want to store it in secret manager (e.g. because ppl scrape github for API keys, etc. etc.)

src/main/java/com/google/sps/servlets/GetElectionServlet.java
```diff
@@ -0,0 +1,169 @@
...
+    Gson gson = new Gson();
+    String json = gson.toJson(elections);
+
+    response.setContentType("application/json");
+    response.getWriter().println(json);
+  }
+
+  // As of now, I kept the doGet nested within a doPut call because I was unsure about whether
+  // a) the doGet method could have a return type that could be passed to the doPut and
+  // b) if the cron job should be calling a doGet when putting something in the Datastore.
```
rogerlaucorp, line 164: Phrase this as a TODO(anooshree): figure out if this should only be doPut() .......

rogerlaucorp, line 164: doPut() would be better if the cloudscheduler is able to. doGet() does not accurately describe what is happening (it's not really getting anything).

rogerlaucorp: If it's not possible, you can just leave a comment stating such.

src/main/java/com/google/sps/servlets/GetElectionServlet.java
```diff
@@ -0,0 +1,169 @@
...
+
+    StringBuilder strBuf = new StringBuilder();
+    HttpURLConnection conn = null;
+    BufferedReader reader = null;
+
+    try {
+      // Running on a local server requires storing API key in environment variable.
+      // In this case, uncomment the following line, and comment out the call to
+      // accessSecretVersion. Do the reverse when deploying.
+      // String electionApiKey = System.getenv("GOOGLE_API_KEY");
```
rogerlaucorp, line 89: This is a pretty low level implementation detail. Maybe have a function getApiKey() that abstracts away which one is happening. Then move the comment above into that functin.

Also something to look into would be if you can control this via a flag or something. Leave a TODO about figuring out via a flag?

src/main/java/com/google/sps/servlets/GetElectionServlet.java
anooshrees: The issue I'm facing is that I need to make a GET request to the API URL to access the newest election information, and I can only do that within a doGet. My question is mainly if I can return the string retrieved from the API from the doGet, pass it to the doPut, and complete the action of putting data in the Datastore within the doPut.

src/main/java/com/google/sps/servlets/ElectionServlet.java
anooshrees: Doing so would require using the cloud version of Datastore (https://cloud.google.com/datastore/docs/concepts/queries) rather than the appengine version we import. This would be primarily a design choice, but would also mean changing a lot of our code because we have been assuming use of the appengine version, which is what we were taught in the tutorials.

rogerlaucorp:
Hey actually Julian tried running this locally. He copied-pasted doGet() into doPut() and it worked fine. Are you sure about the API restriction about only calling GET election from doGet()?

anooshrees:
I ran into issues with the cron job when I tried that, but perhaps that was resolved due to another bug I fixed. I'll try changing it and seeing if it still deploys.

PR #34, opened 2020-07-27
Fix fetching election info from Datastore

casprice:
The code I wrote before wasn't working with Anooshree's latest PR's -- this fixes the bugs.

Deployed already btw - http://contrib-geo-step-2020.uc.r.appspot.com/electionlist.html

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -23,7 +23,7 @@ function listElections() {
     .then((electionList) => {
       // ocdDivisionId is in the form "ocd-division/country:us/state:<state_code>"
       // For example, Lousiana's ID is "ocd-division/country:us/state:la"
-      let regex = /ocd\-division\/country\:us\/(state|district)\:([a-z][a-z])/;
+      let regex = /ocd\-division\/country\:us(\/(state|district)\:([a-z][a-z]))?/;
 
       // For every election returned, store it if the state matches or if national election,
       // to later display the details to the user.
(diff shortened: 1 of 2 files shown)


src/main/webapp/script.js
```diff
@@ -20,28 +20,22 @@ function listElections() {
...
+        let regexMatch = regex.exec(ocdId);
 
-        if (stateMatch != null) {
-          if (stateMatch[1] == selectedStateId) {
-            stateElections.push(election);
-          }
-        } else if (districtMatch != null) {
-          if (districtMatch[1] == selectedStateId) {
+        if (regexMatch != null) {
+          if (regexMatch[2] == selectedStateId) {
```
rogerlaucorp, line 38: combine into:

if (regexMatch != null && regexMatch[2] == selectedStateId) {
} else {
}
?

Should generally try to avoid deep nesting if you can, studies have shown that it greatly reduces readability. Readability is important because studies have also shown the vast majority of time spent in mature products developing is actually spent reading code instead of the actual writing of code. (see 'Clean Code' book for more in depth on those ideas).

casprice: There's an else condition that I wanted to avoid -- if regexMatch[2] != null and regexMatch[2] != selectedStateId.
In that case, we don't want to save it at all, because the state doesn't match. The else condition right now handles if the election is not a state or district election (national election).

There's potentially a different way to handle that, though -- maybe a separate regex specific to national elections so that the else doesn't mistakenly capture a different edge case. I'll make that adjustment!

rogerlaucorp: ACK, ah I see.

src/main/webapp/script.js
```diff
@@ -20,31 +20,23 @@ function listElections() {
...
-            stateElections.push(election);
-          }
-        } else if (districtMatch != null) {
-          if (districtMatch[1] == selectedStateId) {
-            stateElections.push(election);
-          }
-        } else {
+        if (regexMatch != null && regexMatch[3] == selectedStateId) {
+          stateElections.push(election);
+        } else if (regexMatch != null && regexMatch[1] == undefined) {
```
rogerlaucorp, line 39: I like this better too :).

PR #35, opened 2020-07-28
Setup tests

yang-julian:
Sets up basic unit tests and integration tests.

We may want to consider seeding the database with data so that we can get consistent test results.


julian-yang:
I just swapped accounts. I will re open a new PR.

julian-yang:
No idea why Anooshree's changes are getting pulled in here, even though they don't show up in the diffs.. oh well ¯\_(ツ)_/¯

PR #43, opened 2020-07-28
Apply Builder pattern to Election object

casprice:


src/main/java/com/google/sps/data/ElectionBuilder.java
```diff
@@ -0,0 +1,80 @@
+package com.google.sps.data;
+
+import java.util.HashSet;
+
+public class ElectionBuilder {
```
julian-yang, line 5: Please declare this within Election.java.  e.g:
```
public class Election {
  ...
  public static class Builder {
    ...
    build Election build() { }
  } 
}
```

julian-yang: \* OPTIONAL \*

Within Google, we almost always use [AutoValue](https://github.com/google/auto/tree/master/value) for [POJO's](https://en.wikipedia.org/wiki/Plain_old_Java_object).

It makes writing these kind of boiler plate classes a lot easier, but I'm not sure how much setup will be required to get it working.  Essentially you only have to declare which fields you want and the plugin will generate the actual implementing class (getters/setters) for you.  

https://github.com/google/auto/blob/master/value/userguide/index.md 
https://github.com/google/auto/blob/master/value/userguide/builders.md
For example (shamelessly copied from the builders.md link :P):
```
import com.google.auto.value.AutoValue;

@AutoValue
abstract class Animal {
  abstract String name();
  abstract int numberOfLegs();

  static Builder builder() {
    return new AutoValue_Animal.Builder();
  }

  @AutoValue.Builder
  abstract static class Builder {
    abstract Builder setName(String value);
    abstract Builder setNumberOfLegs(int value);
    abstract Animal build();
  }
}
```

casprice: Haha I was just thinking it would be really useful to have something that autogenerates these. Thanks!

rogerlaucorp: +1. At any rate, it's still a useful learning exercise for you to have implemented from scratch haha. This is like a top 5 commonly used design pattern (see: Gang of Four book) that I've seen in the wild. 

I'm okay with you just submitting this if AutoValue is too much work to setup.

PR #48, opened 2020-07-29
Welcome Page and Election List Page testing

gianelgado12:
Added a few integration tests for both the Welcome Page and Election List Page. These tests probe features that are known to work and were done primarily to get familiar with using Selenium for front end testing. The chrome driver necessary to run the tests has also been included under the new resources folder.

### DIFF
--- a/src/test/java/com/google/sps/integration/WelcomePageIT.java
+++ b/src/test/java/com/google/sps/integration/WelcomePageIT.java
@@ -12,8 +12,8 @@ import org.openqa.selenium.chrome.ChromeDriver;
 
 public class WelcomePageIT {
   private WebDriver driver;
-  private final int WAIT_TIME = 30;
-  private final int MAX_NUM_TABS = 2;
+  private static final int WAIT_TIME = 30;
+  private static final int MAX_NUM_TABS = 2;
 
   @BeforeClass
   public static void setUp() {
@@ -30,20 +30,17 @@ public class WelcomePageIT {
    * TurboVotes page.
    */
   @Test
-  public void testTurboVotesButton() {
+  public void turboVotesButton_onClick_redirectsToTurboVotes() {
     driver.get("http://localhost:9876");
 
-    // Ensure only one window is open.
-    assert driver.getWindowHandles().size() == 1;
+    Assert.assertEquals(1, driver.getWindowHandles().size());
 
-    String currentWindow = driver.getWindowHandle();
     WebElement turboVotesButton = driver.findElement(By.id("turbo-vote-button"));
     turboVotesButton.click();
 
-    // Ensure a new tab is created.
     Assert.assertEquals(MAX_NUM_TABS, driver.getWindowHandles().size());
 
-    // Find new tab/window and switch to it.
+    String currentWindow = driver.getWindowHandle();
     for (String windowHandle : driver.getWindowHandles()) {
       if (!currentWindow.contentEquals(windowHandle)) {
         driver.switchTo().window(windowHandle);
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/sps/integration/WelcomePageIT.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.chrome.ChromeDriver;
+
+public class WelcomePageIT {
+  private WebDriver driver;
+  private final int WAIT_TIME = 30;
```
julian-yang, line 15: please make static too

src/test/java/com/google/sps/integration/WelcomePageIT.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Testing if clicking on TurboVotes button opens a new tab and redirects the user to the
+   * TurboVotes page.
+   */
+  @Test
+  public void testTurboVotesButton() {
+    driver.get("http://localhost:9876");
+
+    // Ensure only one window is open.
+    assert driver.getWindowHandles().size() == 1;
```
julian-yang, line 37: Can you use assertEquals() instead?

src/test/java/com/google/sps/integration/WelcomePageIT.java
```diff
@@ -0,0 +1,76 @@
...
+  public void testSetUp() {
+    driver = new ChromeDriver();
+  }
+
+  /**
+   * Testing if clicking on TurboVotes button opens a new tab and redirects the user to the
+   * TurboVotes page.
+   */
+  @Test
+  public void testTurboVotesButton() {
```
julian-yang, line 33: I would recommend making test method names more descriptive.  E.g.
```
turboVotesButton_onClick_redirectsToTurboVotesPage()
```
That way you don't really need the Javadoc _and_ more importantly, when the test fails during a run, you'll know what the failed test case is.

https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

src/test/java/com/google/sps/integration/WelcomePageIT.java
```diff
@@ -0,0 +1,76 @@
...
+    driver.get("http://localhost:9876");
+
+    // Ensure only one window is open.
```
julian-yang, line 36: These comments are not necessary since it's describing _what_ is happening as opposed to _why_.  Generally the code should be already readable where it's easy to tell what's going on by inspection.  There are exceptions to this, but pretty rare.

src/test/java/com/google/sps/integration/WelcomePageIT.java
```diff
@@ -0,0 +1,76 @@
...
+    assert driver.getWindowHandles().size() == 1;
+
+    String currentWindow = driver.getWindowHandle();
```
julian-yang, line 39: please move this down to above the for loop where we use it.  We generally want to keep the declarations as close as possible to their usages.

src/test/java/com/google/sps/integration/WelcomePageIT.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /** Testing if clicking "Show me my elections" button navigates to the elections list */
+  @Test
+  public void testElectionListButton() {
+    driver.get("http://localhost:9876");
+
+    // Ensure only one window is open.
+    assert driver.getWindowHandles().size() == 1;
+
+    String currentWindow = driver.getWindowHandle();
```
julian-yang, line 65: I don't think this is ever used?

PR #49, opened 2020-07-29
Create maven.yml

casprice:
Sets up a Maven workflow on push to master and caches dependencies

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,30 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Java CI with Maven
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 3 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        java-version: 1.8
+    - name: Cache Maven packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml
```
julian-yang, line 30: Hm. is it possible to use `mvn verify` here so that it builds -> unit tests -> runs integration tests? (no idea if we can do this, and we can submit as is if not)

casprice: yeah, I'll add this in

gianelgado12:
@googlebot I fixed it.

PR #53, opened 2020-07-30
Put Election and Candidate info into Datastore

casprice:


src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -98,7 +97,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       String date = (String) entity.getProperty("date");
 
       Election newElection =
-          Election.builder().setID(id).setName(name).setScope(scope).setDate(date).build();
+          Election.builder().setId(id).setName(name).setScope(scope).setDate(date).build();
```
anooshrees, line 100: See my most recent PR -- we need to add the "propositions" and "candidates" sets in order to use the builder; I added empty HashSets for the time being.

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -59,20 +72,77 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       return;
     }
 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity chosenElection = ServletHelper.findElectionInDatastore(datastore, electionId);
+
+    if (chosenElection == null) {
```
anooshrees, line 78: Add this to the ServletHelper class and generalize to other Datastore entity types?

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -59,20 +72,77 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+   */
+  private List<String> fillPositions(DatastoreService datastore, JSONArray positionData) {
+    List<String> positionKeyList = new ArrayList<String>();
+    for (Object positionObject : positionData) {
+      JSONObject position = (JSONObject) positionObject;
+
+      Entity positionEntity = new Entity("Position");
+      positionEntity.setProperty("name", position.getString("office"));
+      positionEntity.setProperty(
+          "candidates", fillCandidates(datastore, position.getJSONArray("candidates")));
```
anooshrees, line 117: So is the "candidates" list a list of Candidate objects or IDs?

src/main/java/com/google/sps/servlets/ElectionServlet.java
```diff
@@ -59,24 +59,23 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
...
        * accidentally reassign IDs to other entities. To avoid this problem, I would have
        * to obtain a block of IDs with allocateIds(), but this is also difficult because
        * election IDs are not always consecutive numbers and other entities we plan to store
        * in Datastore will not have IDs from the Civic Information API (ex. policies) */
-      electionEntity.setProperty("id", currentElection.getLong("id"));
-      electionEntity.setProperty("name", currentElection.getString("name"));
-      electionEntity.setProperty("scope", currentElection.getString("ocdDivisionId"));
-      electionEntity.setProperty("date", currentElection.getString("electionDay"));
+      Entity electionEntity = new Entity("Election");
+      electionEntity.setProperty("electionId", election.getLong("id"));
```
julian-yang, line 75: Can you change this to more object-oriented programming style?  Basically we want to delegate functions to objects.

For example, instead of declaring all of the code here, we can center all of the code around the AutoValue classes.

E.g. have each AutoValue implement these two methods: (using Election as an example):
```
public static Election fromJSONObject(JSonObject obj) {
  // creates a new Election object by extracting the properties from "obj"
} 

public Entity toEntity() {
  // creates a new Entity and sets the proper properties.
}
```

The pros is then you can hide all of the property names from the Servlet code, and reuse it across different servlets that also use that same entity from Civic Api and/or Datastore.

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -29,16 +35,23 @@
...
-  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-
-    // find the corresponding election and check fields
-    // TODO (anooshree): iterate through Datastore and find the election
-    //                   with the matching electionId
-    // if this election is already populated, we don't need to make another query
-
+  public void doPut(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Read in the user-chosen address and election ID, to be used as parameters to the
+    // voterInfoQuery. If these parameters aren't found, API call can't be performed so return
```
julian-yang, line 53: I think you can move this in the javadoc, where you talk about @param request; and then the rest of comment is more of a "what is going on" kind of comment, which we want to avoid

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -59,20 +72,77 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    JSONArray positionListFromApi = electionListFromApi.getJSONArray("contests");
+    chosenElection.setProperty("positions", fillPositions(datastore, positionListFromApi));
+  }
+
+  /**
+   * Puts candidate Position Entities in the Datastore.
+   *
+   * @param datastore the Datastore containing all election data
+   * @param positionData API-returned list of positions on the election ballot
+   * @return list of Key names of all Position Entities added
```
julian-yang, line 107: nit: keys

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -59,20 +72,77 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
         new URL(
             String.format(
+                "%s?address=%s&electionId=%s&key=%s",
                 BASE_URL,
                 address,
                 electionId,
                 ServletHelper.getApiKey("112408856470", "election-api-key", "1")));
-    JSONObject obj = ServletHelper.readFromApiUrl(url, response);
+    JSONObject electionListFromApi = ServletHelper.readFromApiUrl(url);
+    JSONArray positionListFromApi = electionListFromApi.getJSONArray("contests");
```
julian-yang, line 98: Is it possible to keep the same naming scheme as Civic API so that we don't need to keep track of two different names of the same thing?

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -59,20 +72,77 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    chosenElection.setProperty("positions", fillPositions(datastore, positionListFromApi));
```
julian-yang, line 99: I would pull fillPositions() out to a separate line, so that it's a little more clear that you're actually setting datastore keys pointing to the positions rather than the actual position data itself.

src/main/java/com/google/sps/servlets/ServletHelper.java
```diff
@@ -90,4 +92,24 @@ public static JSONObject readFromApiUrl(URL url, HttpServletResponse response)
...
   }
+
+  /**
+   * Find an Election Entity in the Datastore based off its electionId property.
+   *
+   * @param datastore the Datastore containing all election data
+   * @param electionId the ID of the election being queried
+   * @return the Election Entity if found, null otherwise
+   */
+  public static Entity findElectionInDatastore(DatastoreService datastore, String electionId) {
```
julian-yang, line 103: please add @Nullable annotation. (java.util.Nullable)

src/main/java/com/google/sps/servlets/ServletHelper.java
```diff
@@ -43,8 +46,7 @@ public static String getApiKey(String projectId, String secretId, String version
     }
   }
 
-  public static JSONObject readFromApiUrl(URL url, HttpServletResponse response)
-      throws IOException {
+  public static JSONObject readFromApiUrl(URL url) throws IOException {
```
anooshrees, line 49: Why did you choose to remove the response from the parameter list -- how are you alerting the user of back-end errors?

src/main/java/com/google/sps/servlets/ElectionServlet.java
casprice: Yeah, I'll look into this

src/main/java/com/google/sps/servlets/ElectionServlet.java
casprice: Will update!

src/main/java/com/google/sps/servlets/InfoCardServlet.java
casprice: It's the list of ID's. Julian also pointed out this is confusing, so I'll restructure to make it more clear

src/main/java/com/google/sps/servlets/ServletHelper.java
casprice: I removed it because it wasn't being used haha but I agree it should be used to return error messages

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -59,20 +72,77 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    // TODO(anooshree, caseyprice): process JSON objects from API to store as
-    //                              Datastore entities using decomposed functions
+  /**
+   * Puts Candidate Entities in the Datastore.
+   *
+   * @param datastore the Datastore containing all election data
+   * @param candidateData API-returned list of candidates for a position on the election ballot
+   * @return list of Key names of all Candidate Entities added
+   */
+  private List<String> fillCandidates(DatastoreService datastore, JSONArray candidateData) {
```
rogerlaucorp, line 133: Consider renaming this function. "fillCandidates" implies that you're populating some object. Don't be afraid of long but descriptive function names, especially in Java. 

populateDatastoreWithCandidateEntities()?

src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -33,10 +34,31 @@ public static Builder builder() {
...
+  public Entity toEntity() {
+    Entity entity = new Entity("Candidate");
+    entity.setProperty("name", this.getName());
+    entity.setProperty("partyAffiliation", this.getPartyAffiliation());
+    entity.setProperty("campaignSite", this.getCampaignSite());
+    entity.setProperty("platformDescription", this.getPlatformDescription());
+
+    return entity;
+  }
+
```
casprice, line 59: Would an interface be useful for these Entity objects, where the names of some functions are the same but implementation differs? I made an attempt at refactoring to have each of them implement an interface but decided against it because I didn't find it made any external implementation (in InfoCardServlet or ElectionServlet) easier.

rogerlaucorp: It's good that you considered that option, but yeah it sounds like it's unnecessary for now. If it's not actually useful to have an inheritance structure atm, then it's better to not implement one for now and add the additional complexity for no benefit.

See YAGNI (you ain't gonna need it) principle here:
https://www.martinfowler.com/bliki/Yagni.html

Basically as a general rule, default to not over-engineering/complicating the system for a feature that you don't need yet, and do not know for a fact that you will need. A lot of the times you'll find out it was wasted effort/complexity.

casprice:
@googlebot I fixed it.

casprice:
@googlebot please?

anooshrees:
@googlebot  I fixed it.

casprice:
@googlebot I fixed it.

PR #55, opened 2020-07-31
Handle conversion of JSON objects to Datastore Entities

casprice:


src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -33,10 +34,31 @@ public static Builder builder() {
...
+  public Entity toEntity() {
+    Entity entity = new Entity("Candidate");
+    entity.setProperty("name", this.getName());
+    entity.setProperty("partyAffiliation", this.getPartyAffiliation());
+    entity.setProperty("campaignSite", this.getCampaignSite());
+    entity.setProperty("platformDescription", this.getPlatformDescription());
+
+    return entity;
+  }
+
```
casprice, line 59: Would an interface be useful for these Entity objects, where the names of functions are the same but implementation differs? I made an attempt at refactoring to have each of them implement an interface but decided against it because I didn't find it made any external implementation (in InfoCardServlet or ElectionServlet) easier.

PR #56, opened 2020-07-31
Try copying local_db.bin for integration test data seeding

julian-yang:


src/test/java/com/google/sps/integration/ExampleIT.java
```diff
@@ -3,26 +3,55 @@
 import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.net.URL;
-import org.junit.Test;
+import org.junit.*;
```
rogerlaucorp, line 6: Try to avoid * if possible. Do you know what you're trying to import specifically here?

julian-yang: Done.  I think Intellij automatically turned it into *;  I suspect it's because I might be importing all of the packages under org.junit.*?

PR #65, opened 2020-08-03
Refactor ElectionServlet and InfoCardServlet to allow testing with mock objects

casprice:
Seems like this PR pulled in commits from another PR..
Changes made: reading from API response --> POJO implementation has been delegated to ServletUtils.java and POJO code. POJOs now read from JSON and convert to Java object, read from Datastore and convert to Java object, and convert from Java object to Datastore entity.


src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,81 +14,166 @@
...
+
+    if (voterInfoQueryData.has("contests")) {
+      JSONArray contestListData = voterInfoQueryData.getJSONArray("contests");
+      for (Object contestObject : contestListData) {
+        JSONObject contest = (JSONObject) contestObject;
+
+        Contest.fromVoterInfoQuery(datastore, contest).putInDatastore(datastore);
+      }
+
+      contestKeyList = ServletUtils.getEntityKeyNameList(datastore, "Contest");
```
julian-yang, line 105: Wouldn't this return the keys for *all* contests? e.g. if you had Election A and Election B, where Election A has contests A1, A2.  Election B has contests B1, B2.

I believe getEntityKeyNameList returns [A1, A2, B1, B2], where we probably only want [A1, A2] for Election A.

Maybe you can just obtain the contestKeyList from the `putInDatastore()` call on ln 102?

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -26,41 +27,53 @@
...
+      throws JSONException {
+    HashSet<String> candidateKeyList = new HashSet<String>();
+
+    if (contestData.has("candidates")) {
+      for (Object candidateObject : contestData.getJSONArray("candidates")) {
+        JSONObject candidate = (JSONObject) candidateObject;
+        Candidate.fromVoterInfoQuery(candidate).putInDatastore(datastore);
+      }
+
+      candidateKeyList = ServletUtils.getEntityKeyNameList(datastore, "Candidate");
```
julian-yang, line 60: See below in Election.java re Contests.  I think the same applies here with Candidates.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,81 +14,166 @@
...
+    }
+
+    // TODO(caseyprice): get values for propositions
+
+    return Election.builder()
+        .setId(election.getId())
```
julian-yang, line 111: Autovalue has this nifty toBuilder() method:
https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-initialize-a-builder-to-the-same-property-values-as-an-existing-value-instance

src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,81 +14,166 @@
...
+   * properties of "voterInfoQueryData". Delegates creating Contest Entities from this election's
+   * list of contests.
+   *
+   * @param election the Election object already populated by fromElectionQuery
+   * @param datastore the Datastore to store this election's list of contests
+   * @param voterInfoQueryData the JSON output of a voterInfoQuery to the Google Civic Information
+   *     API
+   * @return the new Election object
+   */
+  public static Election fromVoterInfoQuery(
```
julian-yang, line 91: I think this method is probably better off being **not** static.  Then you don't need to pass in the Election object, and can instead call it ON the election object.  E.g.
Right now: `Election.fromVoterInfoQuery(election, datastore, voterInfoQueryData)`
Not static: `election.addVoterInfoQueryData(datastore, voterInfoQueryData)`

src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,81 +14,166 @@
...
+        .build();
+  }
+
+  /**
+   * Checks if an Election object has been populated by the output of a voterInfoQuery call from the
+   * Google Civic Information API.
+   *
+   * @return true if contests and propositions contain elements, false otherwise
+   */
+  public boolean isPopulatedByVoterInfoQuery() {
```
julian-yang, line 126: Is an Election object that has empty contest / proposition data considered to be in an invalid state?
If so, is it possible to delay creating an Election object until we have the contest and proposition data also?

Generally it's best to avoid inserting "invalid" or incomplete data into the datastore as much as we can.  Sometimes this isn't possible though, but it makes handling weird asynchronous cases a lot easier.  For example it's probably possible that the website could read the Election data after it's been inserted into Datastore, but before it has its contest data populated.

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -26,41 +27,53 @@
 public abstract class Contest {
   public abstract String getName();
 
-  public abstract HashMap<Long, Candidate> getCandidates();
+  public abstract HashSet<String> getCandidates();
```
julian-yang, line 30: I think someone added the Guava dependency already (some extra externally available Google Java libraries).  Can we switch this to use ImmutableSet instead?  Basically it prevents other people from adding to the Set after its been created. 

https://effectivejava.googleplex.com/third-edition#page=101

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -55,115 +50,30 @@
...
-    return keyNameList;
+    String url =
+        String.format(
+            BASE_URL,
+            optionalAddress.get(),
+            optionalElectionId.get(),
+            ServletUtils.getApiKey(PROJECT_ID, SECRET_MANAGER_ID, VERSION_ID));
+    JSONObject voterInfoData = ServletUtils.readFromApiUrl(url);
+    Election election = optionalElection.get();
+    Election.fromVoterInfoQuery(election, datastore, voterInfoData).putInDatastore(datastore);
```
julian-yang, line 77: I think this will actually insert a duplicate Election entity instead of updating the existing one right?  Can we update instead of creating new Election entities instead?

src/main/java/com/google/sps/data/Election.java
casprice: You're right, I didn't catch that. I'll make that fix!

src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -34,39 +35,37 @@ public static Builder builder() {
...
+    return Candidate.builder()
+        .setName(candidateData.getString("name"))
+        .setPartyAffiliation(candidateData.getString("party"))
         // TODO(caseyprice): get values for campaignSite and platformDescription
         .setCampaignSite("")
         .setPlatformDescription("")
         .build();
   }
 
   // creates a new Entity and sets the proper properties.
```
rogerlaucorp, line 62: Update this comment to something like "converts the candidate into a datastore entity and puts the entity into the given datastore instance."

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -26,41 +27,53 @@
...
+
+    public abstract Builder setDescription(String description);
+
+    public abstract Contest build();
+  }
+
+  // creates a new Contest object by extracting the properties from "contestData"
+  public static Contest fromVoterInfoQuery(DatastoreService datastore, JSONObject contestData)
+      throws JSONException {
+    HashSet<String> candidateKeyList = new HashSet<String>();
```
rogerlaucorp, line 52: Set<String> candidateKeyList = new HashSet<>();

1) Better to use a less specific interface for the declared type. You don't need to care that it's a HashSet specifically after the initial declaration. 
2) As of java7, HashSet can infer the type, so no need to delcare it explicitly. 

https://stackoverflow.com/questions/29696558/set-without-datatype-mentioned

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -26,41 +27,53 @@
...
+    public abstract Builder setName(String name);
+
+    public abstract Builder setCandidates(HashSet<String> candidates);
+
+    public abstract Builder setDescription(String description);
+
+    public abstract Contest build();
+  }
+
+  // creates a new Contest object by extracting the properties from "contestData"
```
rogerlaucorp, line 49: You should really note any side effects that take place from functions (in this case, the contests are inserted into the datastore).

src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,81 +14,166 @@
...
+   * @param election the Election object already populated by fromElectionQuery
+   * @param datastore the Datastore to store this election's list of contests
+   * @param voterInfoQueryData the JSON output of a voterInfoQuery to the Google Civic Information
+   *     API
+   * @return the new Election object
+   */
+  public static Election fromVoterInfoQuery(
+      Election election, DatastoreService datastore, JSONObject voterInfoQueryData)
+      throws JSONException {
+    HashSet<String> contestKeyList = new HashSet<String>();
```
rogerlaucorp, line 94: Same as above regarding HashSet.

src/main/java/com/google/sps/data/Election.java
rogerlaucorp: +1

src/main/java/com/google/sps/data/Election.java
casprice: Yeah, we use Datastore Election entities without contests/propositions filled out for listing elections on the electionlist.html page. It's not necessarily invalid I think, just incomplete. In this case, we check if it's been populated only to avoid calling the API again when it's not necessary.

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -55,115 +50,30 @@
...
+        ServletUtils.findElectionInDatastore(datastore, optionalElectionId.get());
 
-    Optional<Entity> foundElection = ServletUtils.findElectionInDatastore(datastore, electionId);
-    if (!foundElection.isPresent()) {
-      response.setContentType("text/html");
-      response
-          .getWriter()
-          .println(String.format("Could not find election with ID %s in Datastore.", electionId));
-      response.setStatus(400);
+    if (!optionalElection.isPresent() || (optionalElection.get().isPopulatedByVoterInfoQuery())) {
```
rogerlaucorp, line 65: Consider logging this case for debugging? Or even returning an error if you think it's "fatal".

src/main/java/com/google/sps/servlets/ServletUtils.java
```diff
@@ -21,21 +21,47 @@
...
 
+  /**
+   * Get the value of a query parameter to an HTTP request.
+   *
+   * @param request the HTTP request to be serialized
+   * @param response the HTTP response to publish any error messages
+   * @param inputName the query parameter key
+   * @return Optional container that contains either the parameter or null
+   */
+  public static Optional<String> getRequestParam(
```
rogerlaucorp, line 51: You don't need to do it in this PR yet (although usually we would require it as such), but we should really start adding unit tests soon. Usually we have one file per file under test. So soemthing like ServletUtilsTest.java.

src/test/java/com/google/sps/servlets/ElectionServletTest.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.junit.*;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+/** An example unit test. */
```
rogerlaucorp, line 22: Update this comment. Alternatively, don't have a comment at all, it's pretty clear what this file is.

src/main/java/com/google/sps/servlets/ServletUtils.java
casprice: +1 to this. This PR is being merged to the other PR I have open, since it's mostly refactoring and not really adding functionality. I'll be adding tests for ServletUtils to the other PR (sorry if this is confusing)

rogerlaucorp: ACK

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -26,41 +26,59 @@
...
 
     public abstract Contest build();
   }
+
+  // Creates a new Contest object by extracting the properties from "contestData". For each of its
+  // "candidate" properties, creates a new Candidate object and inserts it into the given Datastore
+  // instance.
+  public static Contest fromVoterInfoQuery(DatastoreService datastore, JSONObject contestData)
+      throws JSONException {
+    HashSet<Long> candidateKeyIds = new HashSet<>();
```
rogerlaucorp, line 55: Set<Long>? I think this can be treated as a more generic Set interface after the initialization.

rogerlaucorp: Same below.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,70 +14,50 @@
...
 
   public abstract String getName();
 
   public abstract String getDate();
 
   public abstract String getScope();
 
+  // This Election references a collection of Contest entities in Datastore. This HashSet represents
+  // their Key names.
   public abstract HashSet<Long> getContests();
```
rogerlaucorp, line 38: Should these return Set instead of HashSet? I don't think it's important how the underlying set is implemented from the client's perspective. 

If this is too annoying to cleanup because it's used in too many places, we can leave a TODO.

src/main/java/com/google/sps/data/Contest.java
casprice: I did some research and I found that the way to do this is
`Set<String> unmodifiableSet = Collections.unmodifiableSet(set);`
where `set` is a `HashSet<String>`.
How would this work with the AutoValue abstract methods, though?

src/main/java/com/google/sps/data/Election.java
```diff
@@ -91,4 +71,123 @@ public Entity toEntity() {
...
+   *
+   * @param election the Election object already populated by fromElectionQuery
+   * @param datastore the Datastore to store this election's list of contests
+   * @param voterInfoQueryData the JSON output of a voterInfoQuery to the Google Civic Information
+   *     API
+   * @return the new Election object
+   */
+  public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterInfoQueryData)
+      throws JSONException {
+    HashSet<Long> contestKeyList = this.getContests();
```
rogerlaucorp, line 106: Same thing regarding sets.

src/main/java/com/google/sps/data/Contest.java
rogerlaucorp: Also if we take Julian's suggestion, this would be ImmutableSet instead of Set, but it's the same idea (we shouldn't care later if this is a TreeSet or a HashSet underlying).

src/main/java/com/google/sps/data/Contest.java
rogerlaucorp: An example:
---
@AutoValue
public abstract class Animal {
  public abstract String name();
  public abstract int numberOfLegs();
  public abstract ImmutableSet<String> countries();

  public static Builder builder() {
    return new AutoValue_Animal.Builder();
  }

  @AutoValue.Builder
  public abstract static class Builder {
    public abstract Builder setName(String value);
    public abstract Builder setNumberOfLegs(int value);
    public abstract Builder setCountries(Set<String> value);
    public abstract Builder setCountries(String... value);
    public abstract Animal build();
  }
}

rogerlaucorp: Yikes that did not format well. Let me see if I can find a link.

rogerlaucorp: I think this site has some examples with autovalue with immutableset:
https://www.codota.com/code/java/classes/com.google.auto.value.AutoValue

But the idea is that you store immutableset, and you return immutableset in the actual AutoValue. You take in a regular set in the Builder.

src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -34,29 +35,6 @@ public static Builder builder() {
     return new AutoValue_Candidate.Builder();
   }
 
-  // creates a new Candidate object by extracting the properties from "obj"
-  public static Candidate fromJSONObject(JSONObject obj) throws JSONException {
```
julian-yang, line 38: Sorry I think I should have asked this earlier.. is there a reason why we're renaming the method from "fromJSONObject" to "fromVoterInfoQuery"?  

Same for Contest#fromVoterInfoQuery and Election#fromElectionQuery.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -91,4 +71,123 @@ public Entity toEntity() {
...
+   * properties of "voterInfoQueryData". Delegates creating Contest Entities from this election's
+   * list of contests.
+   *
+   * @param election the Election object already populated by fromElectionQuery
+   * @param datastore the Datastore to store this election's list of contests
+   * @param voterInfoQueryData the JSON output of a voterInfoQuery to the Google Civic Information
+   *     API
+   * @return the new Election object
+   */
+  public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterInfoQueryData)
```
julian-yang, line 105: Apologies if I gave you the name Election#fromVoterInfoQuery.  I think maybe "fillContestInfo()" might be more descriptive?  IMHO "voter info query" is misleading since it actually doesn't really have anything to do with info about the voter him/herself.

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -55,115 +51,45 @@
    */
   @Override
   public void doPut(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
julian-yang, line 53: Just curious, are we expecting to have the cron job call this endpoint? If so, would it be possible to combine the two calls together?  I think that way we won't need to store Civic api provided ID's in our datastore anymore too, to reduce confusion.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -91,4 +71,123 @@ public Entity toEntity() {
...
+
+  /**
+   * Assigns the given Entity the properties of this Election object and puts it into the given
+   * Datastore instance.
+   *
+   * @param datastore the DatastoreService to store the new Entity
+   */
+  public long putInDatastore(DatastoreService datastore, Entity entity) {
+    /* The "id" of an Election Entity is stored as a property instead of replacing the
+     * Datastore-generated ID because Datastore may accidentally reassign IDs to other
```
julian-yang, line 180: tbh I think all of this can be avoided if we just fetch all of the data upfront so that we won't need to store the civic API's election ID counterpart anymore, but we can just leave it as is.  (no change required here)

src/test/java/com/google/sps/servlets/ElectionServletTest.java
```diff
@@ -0,0 +1,110 @@
...
+  public void testElectionPut() throws Exception {
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+    JSONObject json =
+        new JSONObject(
+            "{\"elections\": [{\"id\": \"9999\",\"name\": \"myElection\",\"electionDay\": \"myDate\",\"ocdDivisionId\": \"myScope\"}]}");
+    JSONArray electionQueryArray = json.getJSONArray("elections");
+
+    for (Object o : electionQueryArray) {
+      JSONObject election = (JSONObject) o;
+      long l = Election.fromElectionQuery(election).addToDatastore(ds);
```
julian-yang, line 76: you can omit this if we're not planning on using it.

src/test/java/com/google/sps/servlets/ElectionServletTest.java
```diff
@@ -0,0 +1,110 @@
...
+    verify(printWriter)
+        .println(
+            "[{\"id\":\"9999\",\"name\":\"myElection\",\"date\":\"myDate\",\"scope\":\"myScope\",\"contests\":[],\"propositions\":[]}]");
+  }
+
+  // Test putting API JSON response for one election in the Datastore and reading from it.
+  @Test
+  public void testElectionPut() throws Exception {
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+    JSONObject json =
```
julian-yang, line 69: nit: electionJson?

src/test/java/com/google/sps/servlets/ElectionServletTest.java
```diff
@@ -0,0 +1,110 @@
...
+    ElectionServlet electionServlet = new ElectionServlet();
+    electionServlet.doGet(httpServletRequest, httpServletResponse);
+    verify(printWriter)
+        .println(
+            "[{\"id\":\"9999\",\"name\":\"myElection\",\"date\":\"myDate\",\"scope\":\"myScope\",\"contests\":[],\"propositions\":[]}]");
+  }
+
+  // Test putting API JSON response for one election in the Datastore and reading from it.
+  @Test
+  public void testElectionPut() throws Exception {
```
julian-yang, line 67: I don't quite understand this test.. are we supposed to be testing the "put" code path also?

src/main/java/com/google/sps/servlets/ServletUtils.java
```diff
@@ -99,20 +124,21 @@ public static JSONObject readFromApiUrl(URL url) throws IOException {
...
     Query query = new Query("Election");
     PreparedQuery results = datastore.prepare(query);
-    Entity targetElection = null;
+    Entity targetEntity = null;
 
     for (Entity entity : results.asIterable()) {
-      if (entity.getProperty("id") == electionId) {
-        targetElection = entity;
+      if (entity.getProperty("id").equals(electionId)) {
+        targetEntity = entity;
```
julian-yang, line 142: can you early return here instead (return Optional.of(entity))

And then at the bottom just return Optional.absent()

src/test/java/com/google/sps/servlets/InfoCardServletTest.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.junit.*;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+public class InfoCardServletTest {
```
julian-yang, line 23: I think this should really be ElectionTest inside of j/c/g/sps/data/?

Seems like you are actually testing Election data class functionality rather than InfoCardServlet.

src/test/java/com/google/sps/servlets/InfoCardServletTest.java
```diff
@@ -0,0 +1,79 @@
...
+                + "{\"id\": \"9999\","
+                + "\"name\": \"myElection\","
+                + "\"electionDay\": \"myDate\","
+                + "\"ocdDivisionId\": \"myScope\""
+                + "}]}");
+    JSONArray electionQueryArray = electionQueryJson.getJSONArray("elections");
+    JSONObject electionJson = electionQueryArray.getJSONObject(0);
+    long entityKeyId = Election.fromElectionQuery(electionJson).addToDatastore(ds);
+
+    when(httpServletResponse.getWriter()).thenReturn(printWriter);
```
julian-yang, line 60: Do you still need this?

src/main/java/com/google/sps/data/Candidate.java
casprice: I did that because Election now has fromElectionQuery AND fromVoterInfoQuery, and I wanted to keep it consistent across the other objects (esp. once we add in other APIs that will also be JSON objects filling in separate fields)

src/main/java/com/google/sps/data/Election.java
casprice: Yeah, I was naming them based off of the API queries I'm getting data from.

src/main/java/com/google/sps/servlets/InfoCardServlet.java
casprice: The PUT requests are intended to be called separately, so doPut in ElectionServlet is called on the electionlist.html page, and stores all of the basic info about the list of elections. Then once an election is chosen doPut in InfoCardServlet is called on the electionInfo.html page and stores info about a specific election.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -92,11 +92,10 @@ public static Election fromElectionQuery(JSONObject electionQueryData) throws JS
   }
 
   /**
-   * Creates a new Election object by extracting the properties from an Election object and from the
```
julian-yang: "Creates" is actually more accurate because toBuilder actually creates a new object. 

https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-initialize-a-builder-to-the-same-property-values-as-an-existing-value-instance
```
... but it would be convenient if they could easily get a Builder instance representing the same 
property values, which they could then modify and **use to create a new value instance.**
```

PR #66, opened 2020-08-04
Add continuity between second and third pages so that query string no…

anooshrees:
…w tracks election name and ID, and third page contains the name of the selected election on the second page.


anooshrees:
@gianelgado12 @casprice the functionality of this redirection is completely working, but the formatting of the text on the third page is now a little off -- if we want to fix it for the MVP, feel free to change the CSS. You should now be able to use the query string to place queries to the Civic Information API: the election ID is "electionId" and the user address is "address".

anooshrees:
I'm going to go ahead and clean up this code and add integration testing

src/main/webapp/electionlist.html
```diff
@@ -94,10 +94,16 @@ <h3>Elections coming up in {{state}}</h3>
...
-                <p>{{this.name}}</p>
+                <div id="state-election-name">
+                  <p>{{this.name}}</p>
+                </div>
                 <p><em>{{formatDate this.date}}</em></p>
               </div>
-              <div class="learn-more-button">Learn more</div>
+              <a href="#" 
+                 onclick="redirectToStateElectionPage({{{this.ID}}});return false;" 
+                 id="election-choice-js">
```
julian-yang, line 104: I don't think you can/should have multiple id tags of the same id on a given page.

src/main/webapp/electionlist.html
```diff
@@ -94,10 +94,16 @@ <h3>Elections coming up in {{state}}</h3>
...
               <div class="election-quick-details">
-                <p>{{this.name}}</p>
+                <div id="state-election-name">
+                  <p>{{this.name}}</p>
+                </div>
                 <p><em>{{formatDate this.date}}</em></p>
               </div>
-              <div class="learn-more-button">Learn more</div>
+              <a href="#" 
+                 onclick="redirectToStateElectionPage({{{this.ID}}});return false;"
```
julian-yang, line 103: I'm not that  familiar with JS, do we need to have the "return false" bit?

src/main/webapp/script.js
```diff
@@ -11,6 +11,16 @@ function onElectionListLoad(){
   }
 }
 
+function onElectionInfoLoad(){
```
julian-yang, line 14: Can we split script.js into mulitple js files, where each js files only contains the functions that are needed for each webpage?

src/main/webapp/script.js
```diff
@@ -99,6 +113,35 @@ function logAddressInput(){
...
+
+/**
+ * Redirect to third page and update the query string with election info.
+ */
+function goToElectionPage(electionName, electionId){
+  addQueryParameter("electionName", electionName);
+  addQueryParameter("electionId", electionId);
+
+  var query = window.location.search;
+  window.location.href = "electionInfo.html" + query;
```
julian-yang, line 142: If we're moving to a new webpage, can't the server fill out the election name?  As opposed to having electionName passed in to this method.

src/test/java/com/google/sps/integration/ElectionInfoIT.java
```diff
@@ -0,0 +1,33 @@
+package com.google.sps.integration;
+
+import org.junit.*;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.chrome.ChromeOptions;
+
+public class ElectionInfoIT {
```
julian-yang, line 8: Can we leave out this file for now since there are no tests inside of it?

anooshrees: yup -- I'll go ahead and remove it now.

src/main/webapp/script.js
anooshrees: How would the server do that?

src/main/webapp/script.js
anooshrees: Sure -- I can make that change right now.

src/main/webapp/electionlist.html
anooshrees: This was how I reasoned we could access the ID from within the handlebars script; is there another way you recommend accomplishing this?

julian-yang: I was referring to 
```
id="election-choice-js"
```

Can't you just change this to `class="..."`?

src/main/webapp/script.js
julian-yang: You probably will want to change electionInfo.html to be dynamically populated.  There are a lot of different ways to do this, using a templating engine is probably the way to go (similar to handlebars.js but done on the server side).

I would recommend using jsp since we're using java servlets.

https://happycoding.io/tutorials/java-server/jsp

I think this tutorial should have most of if not all the information to get started.

anooshrees: Got it -- I had used this method because I was most familiar with it and wanted to complete this task quickly for the sake the MVP, but can switch over to a java servlet soon (later this week or early next week).

src/main/webapp/electionlist.html
anooshrees: oh I see -- will do!

src/main/webapp/script.js
julian-yang: Can you just leave TODO here then?

src/main/webapp/electionlist.html
anooshrees: This keeps the page from redirecting should there be an issue with the first function call (https://stackoverflow.com/questions/128923/whats-the-effect-of-adding-return-false-to-a-click-event-listener).

src/main/webapp/script.js
anooshrees: Definitely -- will do right now. Thanks!

src/test/java/com/google/sps/integration/ElectionListIT.java
```diff
@@ -49,6 +49,29 @@ public void stateSelection_onSelection_createAndUpdateStateQueryParameter() {
...
+    driver.get("http://localhost:9876/electionlist.html?state=ga");
+
+    WebElement learnMoreButton = driver.findElement(By.id("state-learn-more-button"));
+    learnMoreButton.click();
+
+    String targetElectionName = "Georgia+General+Primary+Runoff+Election";
+    String targetElectionID = "4979";
+
+    Assert.assertEquals(
+        "http://localhost:9876/electionInfo.html?state=ga&electionName="
```
rogerlaucorp, line 68: Generally speaking you really wanted to reduce complexity and maximize readability in tests, even more so than regular code. This is important because you want tests to be really obviously correct - or else you'll need tests for your tests (which we sometimes do have...). 

Inline the variables targetElectionName and targetElectionID to maximize easy readability. 

See:
https://drive.google.com/file/d/15GGGANZoaG9JIyGyrTkZWkmYlaih01MJ/view?usp=sharing
On readability in tests.

src/test/java/com/google/sps/integration/ElectionInfoIT.java
```diff
@@ -0,0 +1,33 @@
+package com.google.sps.integration;
+
+import org.junit.*;
```
rogerlaucorp, line 3: Don't use * in import per google style guide. 

The reason why is that it's harder to refactor, less efficient, and less convenient for readers (who often want to see where something is imported from).

src/test/java/com/google/sps/integration/ElectionInfoIT.java
```diff
@@ -0,0 +1,33 @@
+package com.google.sps.integration;
```
rogerlaucorp, line 1: Adding a skeleton file with no tests (yet) is fine, but it'd be nice to have put this in its own PR. Prefer small PRs because they're easier to review. 

See:
https://drive.google.com/file/d/1CNUwfkvB00ZPeAb7aVssF9mIYINzumeZ/view?usp=sharing

src/test/java/com/google/sps/integration/ElectionInfoIT.java
anooshrees: This was done by the auto-compiler, fixing now.

src/test/java/com/google/sps/integration/ElectionInfoIT.java
anooshrees: Good to know, thanks for linking the doc :)

src/test/java/com/google/sps/integration/ElectionListIT.java
anooshrees: Assuming you mean inlining like so:

String string = String.format("A String %s %2d", aStringVar, anIntVar);

src/main/webapp/electionlist.html
```diff
@@ -94,10 +96,16 @@ <h3>Elections coming up in {{state}}</h3>
             {{#each stateElections}}
             <li class="list-group-item election-item">
               <div class="election-quick-details">
-                <p>{{this.name}}</p>
+                <div id="state-election-name">
```
julian-yang, line 99: I think we need to change this "id" too right?, same with "national-election-name"

anooshrees: why would we have to change it? This is how I access the election name for the query URL without mixing up the two types of elections.

julian-yang: Couldn't there be multiple different state elections? you wrapped this in {{#each stateElection}}
Each id="<value>" is supposed to be unique.

anooshrees: oh, I see what you mean. I was under the impression that referring to the div with that id would only capture that particular election because it's within the Handlebars framework.

anooshrees: Let me see if I can pass the name of the object to the function as a parameter instead.

julian-yang: I recommend testing the JS where there are multiple state and national elections to double check that everything works correctly :)

src/main/webapp/election-list.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+  updateLinksWithQueryParams(document.links);
+}
+
+function findElectionNameById(electionId) {
+  fetch('/election')
+    .then(response => response.json())
+    .then((electionList) => {
+      electionList.forEach((election) => {
+        // It's worth noting that, on this branch, the
```
julian-yang, line 72: Just wanted to note that this is not that efficient (fetching every single election in datastore and then parsing in JS).

I'm ok with it since we should be moving to JSP instead soon.
Some other ways to get around this would be to add support for query parameters on /election GET to filter by electionId etc.  (that way the fetch call only returns 1 election).

anooshrees: I see! Let me look into the query parameters for fetch; I'll move to JSP next week for sure!

PR #72, opened 2020-08-05
Test util functions in ServletUtils.java

casprice:


src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+  // Test ServletUtils getRquestParam() function correctly returns empty Optional container when
+  // parameter key does not exist.
+  @Test
+  public void testHTTPRequestGetInvalidParameter() throws Exception {
+    when(httpServletResponse.getWriter()).thenReturn(printWriter);
+
+    Optional<String> parameter =
+        ServletUtils.getRequestParam(httpServletRequest, httpServletResponse, "anInvalidKey");
+    Assert.assertTrue(!parameter.isPresent());
```
julian-yang, line 63: I forget, is there an assertFalse method? If so can we use that instead so we don't need to negate isPresent()?

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+    electionEntity.setProperty("name", "myElection");
+    electionEntity.setProperty("scope", "myScope");
+    electionEntity.setProperty("date", "myDate");
+    electionEntity.setProperty("contests", new HashSet<Long>());
+    electionEntity.setProperty("propositions", new HashSet<Long>());
+    ds.put(electionEntity);
+
+    Optional<Entity> foundEntity = ServletUtils.findElectionInDatastore(ds, "9999");
+    Assert.assertTrue(foundEntity.isPresent());
+    Assert.assertEquals(foundEntity.get().getProperty("name"), "myElection");
```
julian-yang, line 94: Just curious, would assertEquals(foundEntity.get(), electionEntity) fail here?

src/test/java/com/google/sps/unit/ServletUtilsTest.java
casprice: Whoops, there is. Switching to assertFalse for these

src/test/java/com/google/sps/unit/ServletUtilsTest.java
casprice: That works 😅

src/main/java/com/google/sps/servlets/ServletUtils.java
```diff
@@ -49,15 +49,17 @@
   public static Optional<String> getRequestParam(
       HttpServletRequest request, HttpServletResponse response, String inputName)
       throws IOException {
+
     String input = request.getParameter(inputName);
```
rogerlaucorp, line 53: This is not related to this PR, I think I commented before but it got lost in the shuffle. This looks like it's just a collection of static utility functions. Please add a private constructor with a comment
// Private constructor to prevent instantiation. 

See:
http://go/ej3e#page=40

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+  public void setUp() {
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  // Test ServletUtils getRquestParam() function gets parameter in Optional container correctly.
```
rogerlaucorp, line 44: nit: getRequestParam()

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+    Optional<String> parameter =
+        ServletUtils.getRequestParam(httpServletRequest, httpServletResponse, "anInvalidKey");
+    Assert.assertFalse(parameter.isPresent());
+    verify(printWriter).println("No anInvalidKey in the query URL.");
+  }
+
+  // Test ServletUtils getRquestParam() function correctly returns empty Optional container when
+  // provided parameter key is null.
+  @Test
+  public void testHTTPRequestGetNullParameter() throws Exception {
```
rogerlaucorp, line 70: Per google style guide 'HTTP' should still be 'Http' for proper camel case.

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+    when(httpServletResponse.getWriter()).thenReturn(printWriter);
+
+    Optional<String> parameter =
+        ServletUtils.getRequestParam(httpServletRequest, httpServletResponse, null);
```
rogerlaucorp, line 74: This is optional:

To make things more clear, sometimes people will comment what the parameter name is, when the parameter value does not make itself clear (e.g. with null values). 

This is more common in C++, but I see it in Java sometimes too. Not as important in unit tests, but just FYI. 

ServletUtils.getRequestParam(httpServletRequest, httpServletResponse, /*inputName=*/null);

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+    Assert.assertFalse(parameter.isPresent());
+    verify(printWriter).println("No null in the query URL.");
+  }
+
+  // Test finding an Election Entity in Datastore
+  @Test
+  public void testFindExistingElectionInDatastore() throws Exception {
```
rogerlaucorp, line 81: Usually test names should include the name of the function under test. I.e. "findElectionInDatastore" should word-for-word be in the test name. This helps when people are just looking at the overall file structure.

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -0,0 +1,113 @@
...
+    electionEntity.setProperty("id", "9999");
+    electionEntity.setProperty("name", "myElection");
+    electionEntity.setProperty("scope", "myScope");
+    electionEntity.setProperty("date", "myDate");
+    electionEntity.setProperty("contests", new HashSet<Long>());
+    electionEntity.setProperty("propositions", new HashSet<Long>());
+    ds.put(electionEntity);
+
+    Optional<Entity> foundEntity = ServletUtils.findElectionInDatastore(ds, "0001");
+    Assert.assertFalse(foundEntity.isPresent());
```
rogerlaucorp, line 111: (This is FYI only, no action required):

I would recommend using Truth for fluent assertions if this was google internal code:
https://truth.dev/

Which generally makes assertions more readable. But we can skip it for now since it's a bit more setup for google external.

src/test/java/com/google/sps/unit/ServletUtilsTest.java
```diff
@@ -64,21 +64,21 @@ public void testHTTPRequestGetInvalidParameter() throws Exception {
     verify(printWriter).println("No anInvalidKey in the query URL.");
   }
 
-  // Test ServletUtils getRquestParam() function correctly returns empty Optional container when
+  // Test ServletUtils getRequestParam() function correctly returns empty Optional container when
   // provided parameter key is null.
   @Test
-  public void testHTTPRequestGetNullParameter() throws Exception {
+  public void testGetRequestParamWithNullParameterName() throws Exception {
```
julian-yang, line 70: nice, I like the extra descriptions

Can you change it to the format `methodUnderTest_condition1_condition2_expectedOutput`? I don't care which format we use much, but we should keep the naming scheme consistent across our codebase :) 

Similar to https://github.com/googleinterns/step190-2020/blob/master/src/test/java/com/google/sps/integration/ElectionListIT.java

casprice:
Adding onto this PR: I'm adding a function that converts Datastore-produced ArrayList<Long> to HashSet<Long>

casprice:
@rogerlaucorp whoops, just realized I still need your review approval

PR #73, opened 2020-08-06
Contest servlet

gianelgado12:
Added servlet that based on the election Id provided by a query parameter returns all the contests for that election.


src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * This servlet is used to retrieve the information of all the contest present on the election
+ * specified by query parameter.
+ */
+@WebServlet("/contests")
+public class ContestsServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String electionId = ServletUtils.getRequestParam(request, response, "electionId").get();
```
julian-yang, line 42: Can you handle the case where getRequestParam returns Optional.empty()  and can you pull in Casey's changes in  #72?

Same for findElectionInDatastore() (handling Optional.empty())

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+    contestEntityOne.setProperty("name", "myFirstContest");
+    contestEntityOne.setProperty("candidates", candidateIdSet);
+    contestEntityOne.setProperty("description", "This contest is important.");
+
+    contestEntityTwo = new Entity("Contest", 2);
+    contestEntityTwo.setProperty("name", "mySecondContest");
+    contestEntityTwo.setProperty("candidates", new HashSet<Long>());
+    contestEntityTwo.setProperty("description", "This contest is also important.");
+
+    helper.tearDown();
```
julian-yang, line 65: Did you mean to put this here?

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+    electionEntityOne.setProperty("propositions", new HashSet<Long>());
+
+    contestEntityOne = new Entity("Contest", 1);
+    contestEntityOne.setProperty("name", "myFirstContest");
+    contestEntityOne.setProperty("candidates", candidateIdSet);
+    contestEntityOne.setProperty("description", "This contest is important.");
+
+    contestEntityTwo = new Entity("Contest", 2);
+    contestEntityTwo.setProperty("name", "mySecondContest");
+    contestEntityTwo.setProperty("candidates", new HashSet<Long>());
```
julian-yang, line 62: This code should probably be in the setup method annotated @Before instead of @BeforeClass (@BeforeClass is only run once at the before running ALL the tests in this file, where as @Before is run before *each* test)

You shouldn't have to reset contests and propositions entities in @TearDown

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+  private static Entity contestEntityOne;
+  private static Entity contestEntityTwo;
+
+  @Rule public MockitoRule mockitoRule = MockitoJUnit.rule();
+
+  @BeforeClass
+  public static void classSetUp() {
+    helper.setUp();
+
+    candidateIdSet = new HashSet<Long>();
```
julian-yang, line 42: You can make this private static final and use ImmutableSet<Long> instead to declare it all one line.

E.g. ImmutableSet.builder().add(...).add(...).build();

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+public class ContestServletTest {
+  private static final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(
+          new LocalDatastoreServiceTestConfig(), new LocalURLFetchServiceTestConfig());
+
+  @Mock HttpServletRequest httpServletRequest;
+  @Mock HttpServletResponse httpServletResponse;
+  @Mock PrintWriter printWriter;
+
+  private static Set<Long> candidateIdSet;
```
julian-yang, line 31: I think all of these should be final.

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+  @BeforeClass
+  public static void classSetUp() {
+    helper.setUp();
+
+    candidateIdSet = new HashSet<Long>();
+    candidateIdSet.add(Long.valueOf(1));
+    candidateIdSet.add(Long.valueOf(2));
+    candidateIdSet.add(Long.valueOf(3));
+
+    electionEntityOne = new Entity("Election");
```
julian-yang, line 47: I would initialize these in a static block, and then just copy them using `new Entity().setPropertiesFrom(...)` when you want to use them in the tests.

https://www.geeksforgeeks.org/g-fact-79/

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -42,7 +43,7 @@ public static Builder builder() {
   public abstract static class Builder {
     public abstract Builder setName(String name);
 
-    public abstract Builder setCandidates(Set<Long> candidates);
+    public abstract Builder setCandidates(Collection<Long> candidates);
```
rogerlaucorp, line 46: Why are we changing this to Collection? I think we want to leave it as a Set, or else there may be bugs involving repeated candidate ids, no? (Maybe autovalue is smart enough to convert it correctly, but we shouldn't depend on that).

src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * This servlet is used to retrieve the information of all the contest present on the election
```
rogerlaucorp, line 35: nit: contests plural, not contest.

src/main/java/com/google/sps/servlets/ContestsServlet.java
rogerlaucorp: Yeah calling .get() without checking if the optional was present kind of defeats the purpose of optional.

src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+@WebServlet("/contests")
+public class ContestsServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String electionId = ServletUtils.getRequestParam(request, response, "electionId").get();
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity electionEntity;
+
+    electionEntity = ServletUtils.findElectionInDatastore(datastore, electionId).get();
```
rogerlaucorp, line 46: Same here regarding optional. Bad things will happen if you call .get() and it was actually not present. In those cases you want to "gracefully fail" and not give the user some confusing message about how you dereferenced a null optional.

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.appengine.tools.development.testing.LocalURLFetchServiceTestConfig;
+import java.io.PrintWriter;
+import java.util.HashSet;
+import java.util.Set;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.junit.*;
```
rogerlaucorp, line 17: Don't use * per the java style guide. 

The reason why is that it makes refactoring harder, and it makes auto-import/cleanup unused import tools harder, and it makes readability harder (sometimes I want to see where a class is imported from, and * stops me from just ctrl-fing it).

src/test/java/com/google/sps/servlets/ContestServletTest.java
rogerlaucorp: +1.  BeforeClass should be relatively rare usage.

src/test/java/com/google/sps/servlets/ContestServletTest.java
rogerlaucorp: You can also just initialize it directly in the unit test. Having too much setup code hurts readability, which is extra important in tests. Err on the side of DAMP not DRY:

https://drive.google.com/file/d/15GGGANZoaG9JIyGyrTkZWkmYlaih01MJ/view?usp=sharing
(let me know if you don't have permission for this link)

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,122 @@
...
+    verify(printWriter)
+        .println(
+            "[{\"name\":\"myFirstContest\",\"candidates\":[1,2,3],\"description\":\"This contest is important.\"},"
+                + "{\"name\":\"mySecondContest\",\"candidates\":[],\"description\":\"This contest is also important.\"}]");
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+    electionEntityOne.setProperty("contests", new HashSet<Long>());
```
rogerlaucorp, line 119: If you create electionentityone in the each unit test, there's no need to clean this stuff up. Which is nice because it hurts readability.

src/main/java/com/google/sps/data/Contest.java
gianelgado12: I ran into a bug where getProperty in ContestServlet was returning an ArrayList that couldn't be casted to a Set so I decided to use Collections instead as a sort of catch all. Here are the errors I ran into while using HashSet instead of Collection:
Screenshot 2020-08-06 at 11 27 42 AM - Display 1
Screenshot 2020-08-06 at 11 27 42 AM - Display 2

julian-yang: Alternatively you can create a set from the ArrayList that's returned, e.g. `ImmutableSet.copyOf(contestEntity.getProperty(...))`

src/main/webapp/election-info.js
```diff
@@ -22,4 +22,7 @@ function logAddressInput(){
 
   addQueryParameter("address", 
                     `${streetNumber} ${route} ${city} ${state} ${zipCode} ${country}`);
-}
\ No newline at end of file
+
```
rogerlaucorp, line 25: Nit: remove this added whitespace

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+      new LocalServiceTestHelper(
+          new LocalDatastoreServiceTestConfig(), new LocalURLFetchServiceTestConfig());
+
+  @Mock HttpServletRequest httpServletRequest;
+  @Mock HttpServletResponse httpServletResponse;
+  @Mock PrintWriter printWriter;
+
+  private static final Set<Long> candidateIdSet =
+      new ImmutableSet.Builder<Long>()
+          .add(Long.valueOf(1))
```
rogerlaucorp, line 38: Are you sure that you need Long.valueOf() here? I think java autoboxing is smart enough to figure it out. Or maybe if you use 1L, 2L, 3L it'll be smart enough to figure it out.

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+          .add(Long.valueOf(2))
+          .add(Long.valueOf(3))
+          .build();
+  private static Entity electionEntityOne;
+  private static Entity contestEntityOne;
+  private static Entity contestEntityTwo;
```
rogerlaucorp, line 44: Why are these static?

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+public class ContestServletTest {
+  private static final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(
+          new LocalDatastoreServiceTestConfig(), new LocalURLFetchServiceTestConfig());
+
+  @Mock HttpServletRequest httpServletRequest;
+  @Mock HttpServletResponse httpServletResponse;
+  @Mock PrintWriter printWriter;
+
+  private static final Set<Long> candidateIdSet =
```
rogerlaucorp, line 36: private static final stuff are consts which should be all caps per style guide. 

Rename to CANDIDATE_ID_SET.

rogerlaucorp: Also add a space in between this and other variable to indicate separation between regular member variables and the constants.

src/test/java/com/google/sps/servlets/ContestServletTest.java
rogerlaucorp: (the entities; set<long> is fine).

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+public class ContestServletTest {
+  private static final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(
+          new LocalDatastoreServiceTestConfig(), new LocalURLFetchServiceTestConfig());
+
+  @Mock HttpServletRequest httpServletRequest;
+  @Mock HttpServletResponse httpServletResponse;
+  @Mock PrintWriter printWriter;
+
```
rogerlaucorp, line 35: Make the Set an ImmutableSet<Long> instead. 

Generally speaking, later callers of your variable don't need to care if the thing is a TreeSet or HashSet or what, they just need to know it's a Set. 

But Immutable is different - later callers should really know that the thing is Immutable (so that they don't try and mutate it).

src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    String electionId = electionIdOptional.get();
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    Optional<Entity> electionEntityOptional =
+        ServletUtils.findElectionInDatastore(datastore, electionId);
+
+    if (!electionEntityOptional.isPresent()) {
```
rogerlaucorp, line 59: Add a test case for this condition.

src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -0,0 +1,91 @@
...
+ * specified by query parameter.
+ */
+@WebServlet("/contests")
+public class ContestsServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Optional<String> electionIdOptional =
+        ServletUtils.getRequestParam(request, response, "electionId");
+
+    if (!electionIdOptional.isPresent()) {
```
rogerlaucorp, line 48: Add a test case for this condition. You want your unit tests to have good "coverage" which means that the unit tests exercise every line (or as many lines as possible) of your code. 

There's tool and stuff to do analysis about what your test coverage is, but for now just know that unit tests should aim to maximize code coverage (among other goals).

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,131 @@
...
+    contestEntityOne.setProperty("description", "This contest is important.");
+
+    contestEntityTwo = new Entity("Contest");
+    contestEntityTwo.setProperty("name", "mySecondContest");
+    contestEntityTwo.setProperty("candidates", new HashSet<Long>());
+    contestEntityTwo.setProperty("description", "This contest is also important.");
+  }
+
+  @Test
+  public void singleElection_singleContest_testDoGet() throws Exception {
```
rogerlaucorp, line 71: Add a test case for when there are no contests to exercise "corner cases".

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: That was originally so that they could be referenced from within the static @BeforeClass setup method but its unnecessary now.

src/main/webapp/election-info.js
```diff
@@ -22,4 +22,6 @@ function logAddressInput(){
 
   addQueryParameter("address", 
                     `${streetNumber} ${route} ${city} ${state} ${zipCode} ${country}`);
-}
\ No newline at end of file
+}
+
```
rogerlaucorp, line 26: Remove this random whitespace?

src/main/webapp/election-info.js
gianelgado12: Done

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: Done

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: Done

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: done

src/main/java/com/google/sps/servlets/ContestsServlet.java
gianelgado12: done

src/main/java/com/google/sps/servlets/ContestsServlet.java
gianelgado12: done

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: done

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: made entities non-static

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: Placed entity set up code in setUp method tagged with @Before

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: Removed method tagged @BeforeClass

rogerlaucorp:
@googlebot I fixed it.

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -75,9 +76,9 @@ public static Contest fromJSONObject(DatastoreService datastore, JSONObject cont
   // Creates a new Contest object by using the propperties of the provided
   // contenst entity
   public static Contest fromEntity(Entity entity) {
-    HashSet<Long> candidates = new HashSet<>();
+    Set<Long> candidates = new HashSet<>();
```
julian-yang, line 79: nit: ImmutableSet.of()

Can just leave it as an empty set.

src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -0,0 +1,91 @@
...
+      return;
```
julian-yang, line 49: Can you add an error message here too?  similar to ln 60-63

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+    verify(printWriter)
+        .println(
+            "[{\"name\":\"myFirstContest\",\"candidates\":[1,2,3],\"description\":\"This contest is important.\"}]");
+  }
+
+  @Test
+  public void singleElection_twoContest_testDoGet() throws Exception {
+    Entity electionEntity = new Entity("Election");
+    Entity firstContestEntity = new Entity("Contest", 1);
```
julian-yang, line 96: Do we need to specify the id?

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+    electionEntity.setPropertiesFrom(electionEntityOne);
+    firstContestEntity.setPropertiesFrom(contestEntityOne);
+    secondContestEntity.setPropertiesFrom(contestEntityTwo);
+
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+    long contestOneId = ds.put(firstContestEntity).getId();
+    long contestTwoId = ds.put(secondContestEntity).getId();
+
+    HashSet<Long> contestSet = (HashSet<Long>) electionEntityOne.getProperty("contests");
```
julian-yang, line 107: I thought we couldn't cast into HashSet? maybe you can use a HashSet constructor instead?

Feel free to just resolve if this runs successfully :)

src/main/java/com/google/sps/servlets/ContestsServlet.java
gianelgado12: ServerUtils.getRequestParams already writes an error message to response when it returns an empty Optional which seemed sufficient to me.

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: That was leftover from when we had the setUpClass method but it seems to work fine now without specifying the Id. Will update.

gianelgado12: Actually still need to specify the Id when testing when a contest isn't present in datastore for the entity that is created but isn't put into datastore.

src/main/java/com/google/sps/data/Contest.java
gianelgado12: done

src/test/java/com/google/sps/servlets/ContestServletTest.java
gianelgado12: done

PR #78, opened 2020-08-06
Debugging redirection function so that second and third page connect.

anooshrees:


anooshrees:
The Maven test fails due to an error of "No such element" -- this is because the election list for a state isn't loading because fetch("/elections") doesn't yield anything locally. @casprice ran into the same issue on her local machine, but if /elections contains data, the test works.

PR #79, opened 2020-08-07
Polling station class and servlet

anooshrees:


src/main/java/com/google/sps/data/Election.java
```diff
@@ -42,6 +44,8 @@
   // represents their Key names.
   public abstract Set<Long> getPropositions();
 
+  public abstract ArrayList<EmbeddedEntity> getPollingStations();
```
rogerlaucorp, line 47: Can we return an immutableset here? Store it as an arraylist in datastore, but when you get it out, put it into a set. And make it immutable to reduce bugs. 

See: go/ej3e-17

src/main/java/com/google/sps/data/Election.java
```diff
@@ -140,6 +199,8 @@ public boolean isPopulatedByVoterInfoQuery() {
   public static Election fromEntity(Entity entity) {
     Set<Long> contests = new HashSet<>();
     Set<Long> propositions = new HashSet<>();
+    ArrayList<EmbeddedEntity> pollingStations = new ArrayList<EmbeddedEntity>();
```
rogerlaucorp, line 202: List<EmbeddedEntity> pollingStations = new ArrayList<>();

(1) Java is smart enough to infer the type at runtime. 
(2) Prefer the more general interface.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -116,9 +130,52 @@ public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterI
       }
     }
 
+    if (voterInfoQueryData.has("earlyVoteSites")) {
+      JSONArray earlyVoteSiteData = voterInfoQueryData.getJSONArray("earlyVoteSites");
```
rogerlaucorp, line 134: We should put these variables in to constants at some point.

src/main/java/com/google/sps/data/Election.java
anooshrees: Doing so gives me the following error:

java.util.List<com.google.appengine.api.datastore.EmbeddedEntity> cannot be converted to java.util.ArrayList<com.google.appengine.api.datastore.EmbeddedEntity>

src/main/java/com/google/sps/data/Election.java
anooshrees: Got it! By the way, the go link to specific pages in the book is super cool

anooshrees: Do you mean change it to an immutable set in the Election class itself, or when retrieved through the PollingStationServlet?

src/main/java/com/google/sps/data/Election.java
anooshrees: I can do this right now. I'm thinking of making a constant array consisting of these three types and iterating through that to avoid the code repetition present right now.

anooshrees:
By the way, the blue doc relevant to this PR is available at go/step190-datastore-blue-doc.

src/main/java/com/google/sps/data/Election.java
rogerlaucorp: In the election class itself, and just transform it from an ArrayList when you instantiate this class. i.e. 

public abstract ImmutableSet<EmbeddedEntity> getPollingStations();

See gianluca's similar code here:
https://github.com/googleinterns/step190-2020/blob/master/src/main/java/com/google/sps/data/Contest.java#L81

src/main/java/com/google/sps/data/Election.java
rogerlaucorp: I think that's because you have to change the builder function to also take in a list. And return a list. 

But anyway, combined with the comment from above about immutableset, all this should be converted to sets/immutableset. 

Here:
Set<EmbeddedEntity> pollingStations = new HashSet<>();

In builder:
public abstract Builder setPollingStations(Set<EmbeddedEntity> pollingStations);

In autovalue:
public abstract ImmutableSet<EmbeddedEntity> pollingStations();
----------
The same advice from before applies though. 
 * Set<EmbeddedEntity> in the variable declaration because you want it more generic. 
 * new HashSet<> in the instantiation because Java is smart enough to infer the type
 * ImmutableSet because mutability is bad

src/main/java/com/google/sps/data/Election.java
anooshrees: @rogerlaucorp I put a TODO in my code but per our discussion in the blue doc, I was thinking of using an enum to store these constant values. I know @casprice used static final Strings to hold constant values, but afaik Java can handle enum constants with String values (https://howtodoinjava.com/java/enum/java-enum-string-example/) and therefore this could be a neat way to handle it. My only concern is if enums are compliant with Datastore, which I am currently looking into now.

rogerlaucorp: TODO is fine. If they're not support in datastore (I suspect they're not) you can store an int or string, and then just parse it when you read from datastore.

PR #80, opened 2020-08-07
Get Referendum data

casprice:


src/main/java/com/google/sps/data/Election.java
```diff
@@ -104,31 +104,35 @@ public static Election fromElectionQuery(JSONObject electionQueryData) throws JS
...
     if (voterInfoQueryData.has("contests")) {
       JSONArray contestListData = voterInfoQueryData.getJSONArray("contests");
       for (Object contestObject : contestListData) {
         JSONObject contest = (JSONObject) contestObject;
 
-        long contestEntityKeyId =
-            Contest.fromJSONObject(datastore, contest).addToDatastore(datastore);
-        contestKeyList.add(contestEntityKeyId);
+        // Referendums are a separate contest type, so separate them out from the office positions.
+        if (contest.getString("type").equals("Referendum")) {
```
rogerlaucorp, line 114: Maybe put "referendum" into an enum instead of hard coding? Then we can also use a switch here.

casprice: @rogerlaucorp can enums be given String values? Or would it be better to use a static constant here?

rogerlaucorp: You could do something like this:
https://howtodoinjava.com/java/enum/java-enum-string-example/
or like this:
https://www.programiz.com/java-programming/enum-string

PR #81, opened 2020-08-07
Get Referendum data

casprice:
Remade this PR because the last one was on a messy branch.


src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -23,6 +23,14 @@
 /** A candidate for a public office position that will appear on voter ballots */
 @AutoValue
 public abstract class Candidate {
+  public static final String ENTITY_NAME = "Candidate";
+  public static final String NAME_JSON_KEYWORD = "name";
+  public static final String PARTY_JSON_KEYWORD = "party";
+  public static final String NAME_OBJECT_KEYWORD = "name";
```
julian-yang, line 29: Hm, should we replace this with "ENTITY"? since they should be used in from/to Entity related methods.

src/main/java/com/google/sps/servlets/ContestsServlet.java
```diff
@@ -82,10 +84,37 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    String json = gson.toJson(contestList);
+    String contestJson = gson.toJson(contestList);
+    String referendumJson = gson.toJson(referendumList);
 
     response.setContentType("application/json;");
-    response.getWriter().println(json);
+    response
+        .getWriter()
+        .println(
+            "{\""
```
julian-yang, line 110: I feel like it's kind of confusing that the /contests servlet returns both "contests" and "referendums", but I don't have a better naming suggestion haha.

I think it's worth thinking about for maybe a few minutes, but if you can't think of any better naming schemes, I'm ok with submitting as is.

src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -23,6 +23,14 @@
 /** A candidate for a public office position that will appear on voter ballots */
 @AutoValue
 public abstract class Candidate {
+  public static final String ENTITY_NAME = "Candidate";
```
julian-yang, line 26: maybe ENTITY_TYPE or ENTITY_KIND.

I think "kind" is the nomenclature that Datastore documentation uses.
https://cloud.google.com/datastore/docs/concepts/entities#kinds_and_identifiers

src/main/java/com/google/sps/servlets/ContestsServlet.java
casprice: Hmm to be honest it might make more sense to rename the Contest object, that way it stays consistent with the API

julian-yang: I'm ok with that too :)

PR #87, opened 2020-08-10
Add Candidate info to ContestServlet GET response

casprice:
Serializes Candiate objects corresponding to each Contest and attaches their JSON objects to the "candidates" field of each "contest" in ContestServlet's doGet().

### DIFF
--- a/src/main/java/com/google/sps/data/Candidate.java
+++ b/src/main/java/com/google/sps/data/Candidate.java
@@ -17,6 +17,7 @@ package com.google.sps.data;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.Entity;
 import com.google.auto.value.AutoValue;
+import com.google.gson.Gson;
 import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/sps/data/Contest.java
```diff
@@ -83,6 +93,30 @@ public static Contest fromJSONObject(DatastoreService datastore, JSONObject cont
         .build();
   }
 
+  // Convert this Contest object to a String in JSON format. Serializes its list of Candiate Entity
+  // key names into JSON objects that correspond to each Candidate's attributes.
+  public String toJsonString(DatastoreService datastore) {
+    Gson gson = new Gson();
+    JsonArray candidateJsonArray = new JsonArray();
+
+    // Since a Contest object's "candidates" collection consists only of their Datastore key names,
```
julian-yang, line 102: Did you mean Datastore ids? you call the variable "contestCandidateIds" below :P

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -83,6 +93,30 @@ public static Contest fromJSONObject(DatastoreService datastore, JSONObject cont
...
+    // iterate through each one to get the actual Candidate field data in JSON format.
+    Set<Long> contestCandidateIds = ImmutableSet.copyOf(this.getCandidates());
+    for (Long candidateId : contestCandidateIds) {
```
julian-yang, line 105: \* no action required / FYI * 

I would consider using Java stream here, since you're really doing a series of transformations on the Set.
1) id -> key
2) key -> datastore lookup
3) entity -> json

followed by collecting all of the json strings into the json array.

src/main/java/com/google/sps/data/Contest.java
casprice: Oop that should be "key name" all around, actually. Will update!

casprice: Wait jk it's ID

src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -67,6 +68,21 @@ public static Candidate fromJSONObject(JSONObject candidateData) throws JSONExce
         .build();
   }
 
+  // Converts this Candidate object to a JSON string.
+  public String toJsonString() {
+    return new Gson().toJson(this);
+  }
+
+  // Creates a new Candidate object by using the propperties of the provided Candidate entity
```
julian-yang, line 76: nit: properties

PR #88, opened 2020-08-11
Fix broken Maven build by updating integration_db.bin

casprice:


casprice:
> Just curious, what did you change in integration_db.bin? and what was failing?

No idea, everything passed in the PR I just merged. It failed for some reason on master.
When this happens, I usually just run the code on the local server and do a PUT request to make sure /elections is populated, and then recopy the local_db.bin.

PR #97, opened 2020-08-12
Create UI for MVP.

anooshrees:
Changes made:
* HTML/CSS for the Info cards
* Handlebars template for info cards, to be filled in by Servlet responses
* JS functionality for getting the Election, Contest, Candidate, Referendum, and PollingStation data and populating template
* Candidate and PollingStation objects handle if API JSON fields don't exist
* Functionality for making info cards expandable/collapsible on click
* List of PollingStations added as attribute of Election
* Map displays on electionInfo page to mark polling stations nearby
* Address input submit button is disabled on default and is enabled when all address input fields are filled


rogerlaucorp:
Are the tests broken?

src/main/java/com/google/sps/data/Candidate.java
```diff
@@ -60,10 +60,32 @@ public static Builder builder() {
 
   // Creates a new Candidate object by extracting the properties from "candidateData"
   public static Candidate fromJSONObject(JSONObject candidateData) throws JSONException {
```
rogerlaucorp, line 62: Does this function still throw a JSONException after you've try-catched everything?

rogerlaucorp: Remove "throws JSONException" if not.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -132,8 +133,9 @@ public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterI
       throws JSONException {
     Set<Long> contestKeyList = this.getContests();
     Set<Long> referendumKeyList = this.getReferendums();
-    ArrayList<EmbeddedEntity> pollingStations =
-        new ArrayList<EmbeddedEntity>(this.getPollingStations());
+
+    List<EmbeddedEntity> pollingStations = new ArrayList<EmbeddedEntity>(this.getPollingStations());
```
rogerlaucorp, line 137: You can remove the type from the instantiation now, java got smarter.

new ArrayList<>(this.getPollingStations());

src/main/java/com/google/sps/data/Election.java
```diff
@@ -157,29 +159,36 @@ public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterI
...
       for (Object pollingLocation : pollingLocationData) {
         JSONObject pollingLocationJSON = (JSONObject) pollingLocation;
-        pollingStations.add(createPollingStation(pollingLocationJSON, "pollingLocation"));
+        pollingStations.add(
+            createPollingStation(pollingLocationJSON, "pollingLocation", datastore));
       }
     }
 
+    if (pollingStations.size() > 0) {
+      pollingStationSet = ImmutableSet.copyOf((Collection<EmbeddedEntity>) pollingStations);
```
rogerlaucorp, line 184: Do you need this if condition? a copy of an empty set is still just an empty set, which is what the else is doing anything.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -132,8 +133,9 @@ public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterI
       throws JSONException {
     Set<Long> contestKeyList = this.getContests();
     Set<Long> referendumKeyList = this.getReferendums();
-    ArrayList<EmbeddedEntity> pollingStations =
-        new ArrayList<EmbeddedEntity>(this.getPollingStations());
+
+    List<EmbeddedEntity> pollingStations = new ArrayList<EmbeddedEntity>(this.getPollingStations());
+    ImmutableSet<EmbeddedEntity> pollingStationSet;
```
rogerlaucorp, line 138: Do you know if the EmbeddedEntity hashcode/equals functions behave well for use in sets?

Sets depend on objects implementing equals() and hashcode(). Sometimes they will not behave as you expect (e.g. instead of checking that all the keys and values are the same, maybe they check that the variables literally refer to the same object, like pointer equality). 

See: go/ej3e-10

src/main/java/com/google/sps/data/Election.java
```diff
@@ -191,12 +200,17 @@ public Election fromVoterInfoQuery(DatastoreService datastore, JSONObject voterI
    * @param locationType a string signifying if this location is a polling station, drop-off
```
rogerlaucorp, line 200: Document the datastore param

src/main/java/com/google/sps/data/Election.java
```diff
@@ -235,6 +249,8 @@ public static Election fromEntity(Entity entity) {
     if (entity.getProperty("pollingStations") != null) {
       pollingStationSet =
           ImmutableSet.copyOf((Collection<EmbeddedEntity>) entity.getProperty("pollingStations"));
+    } else {
```
rogerlaucorp, line 252: Same comment as above regarding if-condition.

src/main/java/com/google/sps/data/Election.java
anooshrees: got it, thanks!

src/main/java/com/google/sps/data/Election.java
anooshrees: according to this documentation (https://googleapis.dev/python/datastore/latest/entities.html), "Entities compare equal if their keys compare equal and their properties compare equal" using the __eq__ method. EmbeddedEntities are simply Entities intended to be stored as properties, so I'm assuming that the same functionality extends here, based on the description here (https://cloud.google.com/appengine/docs/standard/java/datastore/entities#Embedded_entities).

src/main/java/com/google/sps/data/Election.java
anooshrees: You're correct -- I was using this line of code for debugging purposes and will delete it now.

src/main/java/com/google/sps/data/Election.java
anooshrees: Got it, will do!

src/main/java/com/google/sps/data/Election.java
anooshrees: My worry here was about instantiating the ImmutableSet as empty then changing its value, but it looks like ImmutableSets only refuse the add() method, so I'll change this.

anooshrees: The issue here is that Datastore stores empty lists as "null", so let me check if I can call copyOf() on a null object. If not, I'll just instantiate pollingStationSet as an empty ImmutableSet at the start of the function.

rogerlaucorp: If not this is fine, actually i misread the if-statement and thought it was checking the siz elike above.

anooshrees:
> Are the tests broken?

No, but we have to update the .db file occasionally to ensure that the test that relies on the localhost's datastore works correctly.

PR #107, opened 2020-08-12
Delete all data from voterInfoQuery on PUT request to avoid repeated data

casprice:


src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -17,9 +17,13 @@
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
-import com.google.sps.data.Election;
+import com.google.common.collect.ImmutableSet;
+import com.google.sps.data.*;
```
rogerlaucorp, line 21: Don't import *

src/main/java/com/google/sps/servlets/ServletUtils.java
```diff
@@ -140,6 +141,16 @@ public static JSONObject readFromApiUrl(String urlString) throws IOException {
     return obj;
   }
 
+  public static void deleteAllEntitiesOfKind(DatastoreService datastore, String entityKind) {
```
rogerlaucorp, line 144: Unit test for this?

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -76,13 +97,16 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
...
+    logger.logp(Level.INFO, SOURCE_CLASS, "doPut", "Got election ID and user address.");
+
     Entity electionEntity = optionalEntity.get();
-    Election election = Election.fromEntity(electionEntity);
 
-    // Don't need to make the API call if this Election object has already been populated.
-    if (election.isPopulatedByVoterInfoQuery()) {
-      return;
-    }
+    // Temporary solution to repeated information in the Datastore.
```
rogerlaucorp, line 104: Leave a TODO for gianluca, we asked him to write a one pager on this and work on it. Also can you elaborate more on the problem in the comment, pls?

src/main/java/com/google/sps/servlets/InfoCardServlet.java
casprice: This is our local data folder, so it only has 5 files (all of which we use). Is using * ok in this situation?

PR #109, opened 2020-08-13
Election info error screen when no Election ID is provided

casprice:
I spent some time looking into writing tests for the changes I made, but it seems Selenium doesn't work well with Handlebars.js templates. I'll try doing JS tests instead tomorrow!
Edit: Scratch JS tests, we decided they don't add to the testing coverage
Edit 2: Got Selenium working, turns out I was just writing them wrong hahah


src/main/webapp/election-info.js
```diff
@@ -1,15 +1,35 @@
...
 
-    let titleTextElement = document.getElementById('election-info-page-title');
-    titleTextElement.innerHTML = template(context);
-  }
+    if (context == null) {
+      context = { 
+        electionIdInQuery: false,
+      };
+      document.getElementById('election-info-wrapper').style.display = 'none';
+      titleTextElement.innerHTML = template(context);
```
julian-yang, line 30: Should this be outside of `if (context == null)`?

src/main/webapp/election-info.js
```diff
@@ -1,15 +1,35 @@
 function onElectionInfoLoad(){
   let searchParams = new URLSearchParams(window.location.search);
```
julian-yang, line 2: Can you push all of these declarations inside the .then()?

PR #110, opened 2020-08-13
Catch JSONException when referendumSubtitle does not exist

gianelgado12:
Deployed site had issues where a 500 error was thrown when the referendumSubtitle field didn't exist since JSONObject.getString() threw an uncaught JSONException. Fixed issue by catching the JSONException and subsequently filling in the description field in Referendum with an empty string.

### DIFF
--- a/src/main/java/com/google/sps/data/Referendum.java
+++ b/src/main/java/com/google/sps/data/Referendum.java
@@ -49,9 +49,24 @@ public abstract class Referendum {
 
   // creates a new Referendum object by extracting the properties from "obj"
   public static Referendum fromJSONObject(JSONObject obj) throws JSONException {
+    String referendumDescription;
+    String referendumTitle;
+
+    try {
+      referendumTitle = obj.getString(TITLE_JSON_KEYWORD);
+    } catch (JSONException e) {
+      throw new JSONException("Referendum title not found in JSON response.");
+    }
+
+    try {
+      referendumDescription = obj.getString(DESCRIPTION_JSON_KEYWORD);
+    } catch (JSONException e) {
+      referendumDescription = "";
+    }
+
     return Referendum.builder()
-        .setTitle(obj.getString(TITLE_JSON_KEYWORD))
-        .setDescription(obj.getString(DESCRIPTION_JSON_KEYWORD))
+        .setTitle(referendumTitle)
+        .setDescription(referendumDescription)
         .build();
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/data/Referendum.java
```diff
@@ -49,9 +49,24 @@ public static Builder builder() {
...
+    try {
+      referendumTitle = obj.getString(TITLE_JSON_KEYWORD);
+    } catch (JSONException e) {
+      throw new JSONException("Referendum title not found in JSON response.");
+    }
+
+    try {
+      referendumDescription = obj.getString(DESCRIPTION_JSON_KEYWORD);
+    } catch (JSONException e) {
+      referendumDescription = "";
```
julian-yang, line 64: please add logging.  Why are we throwing an exception above but not here?

rogerlaucorp:
Good comment overview.

src/test/java/com/google/sps/data/ReferendumTest.java
```diff
@@ -45,6 +52,28 @@ public void testAddElectionQueryData() throws Exception {
...
+  }
+
+  @Test
+  public void missingTitle_throwJSONException_testFromJSONObject() throws Exception {
+    JSONObject referendumJsonObject =
+        new JSONObject(
+            "{\"type\": \"Referendum\", \"referendumSubtitle\": \"Water Bond. Funding for Water Quality, Supply, Treatment, and Storage Projects.\"}");
+
+    exceptionRule.expect(JSONException.class);
+    exceptionRule.expectMessage("Referendum title not found in JSON response.");
```
rogerlaucorp, line 73: Prefer assertThrows:
go/java-testing/exceptions#assertThrows
https://screenshot.googleplex.com/4tnrOY4ybCJ

(lemme know if you can't access that site)

rogerlaucorp: There are a few reasons for this, but a big one is that theoretically any of the followign code could throw the exception, and this test would pass, even if it's not the actual line you intend. 

This would probably be okay in this case, but it's easier to just always follow the rule instead of analyzing the code each time to see what you can get away with for no additional benefit.

src/main/java/com/google/sps/data/Referendum.java
gianelgado12: The rationale is that it's possible for a referendum to be missing a referendumSubtitle field and in that case its sufficient to just catch the exception and fill in the description field with an emptyString. However, a missing referendumTitle probably indicates some sort of malformed response from the API or some undesired manipulation of the obj JSONObject which would warrant throwing an exception.  Will add logging.

src/test/java/com/google/sps/data/ReferendumTest.java
gianelgado12: For some reason assertThrows doesn't seem to exist in JUnit4 only in JUnit5

rogerlaucorp: ACK.

PR #111, opened 2020-08-13
Address input IT

anooshrees:
This test will never work with the Github Maven check because it relies on the Places API to work, which in turn relies on the key in our config.js file, which is in .gitignore. That being said, it is successful locally and does not change the functionality of our deployed site.

Also, it looks like this sucked in the commits from another branch -- check out the last commit for the actual changes.


src/test/java/com/google/sps/integration/ElectionInfoIT.java
```diff
@@ -25,9 +32,46 @@ public void testSetUp() {
...
+    driver.get("http://localhost:9876/electionInfo.html?state=ca&electionId=2000");
+    WebElement submitButton = driver.findElement(By.id("submit-address-button"));
+    submitButton.click();
+
+    Assert.assertEquals(
+        "http://localhost:9876/electionInfo.html?state=ca&electionId=2000", driver.getCurrentUrl());
+  }
+
+  @Test
+  public void addressSubmission_onClick_createAndUpdateQueryParams() throws InterruptedException {
```
julian-yang, line 49: Sorry I think I misunderstood you earlier, I don't think it makes sense to submit this test if it can't pass maven verify.

I don't think the github system will let you merge these changes either?

anooshrees: It passes maven verify, just not on github because config.js isn't pushed to github. I am pretty sure you can push regardless.

rogerlaucorp: You can just delete the test and leave a TODO explaining the problem. We'll punt on it for now, but it's bad to check something in that will permanently break the github presubmit tests.

PR #112, opened 2020-08-13
Loading icon

anooshrees:
Functionality viewable (as of pull time) on the deployed website!


anooshrees:
BTW, @casprice and I realized that the test receiving "NoSuchElement" Exception during the Maven Github check is erroring because the election it's checking -- the Georgia primary -- no longer exists (go cron job!). Casey is changing it in her upcoming PR, but I figured I'd let everyone know why the check is failing here.

PR #114, opened 2020-08-14
Make Candiate Campaign URLs Clickable

gianelgado12:
Now when a candidate has a campaign URL returned by the civic information API the user can click on the URL in the candidates dropdown and the candidate's campaign site will open in a new tab. Also, if a candidate does not have a campaign URL returned by the civic information API that field is not shown at all in the candidate's dropdown. Additionally, the URL displayed in the dropdown is stripped of the trailing slash and preceding 'http://' to clean up the look of the dropdown.


src/main/webapp/election-info.js
```diff
@@ -256,4 +256,9 @@ function addPollingStationMarker(map, position, title, description) {
   marker.addListener('click', () => {
     infoWindow.open(map, marker);
   });
-}
\ No newline at end of file
+}
+
+Handlebars.registerHelper('stripUrl', function(urlString){
```
julian-yang, line 261: Can you add some comments / documentation on what this is supposed to be used for?

I'm guessing it's meant to strip off the "http://" part of the url

gianelgado12: done

src/main/webapp/election-info.js
```diff
@@ -256,4 +256,15 @@ function addPollingStationMarker(map, position, title, description) {
...
+
+/**
+ * Handlebars helper that removes the beginning "http://" and the trailing "/"
+ * from the supplied URL string.
+ * 
+ * @param {String} URL to be stripped.
+ */
+Handlebars.registerHelper('stripUrl', function(urlString){
+  urlString = urlString.replace(/(^\w+:|^)\/\//, '');
+  return urlString.slice(0, -1);
```
julian-yang, line 269: Please check whether the last character is actually "/" instead of blindly dropping the last character?

gianelgado12: done

src/main/webapp/electionInfo.html
```diff
@@ -154,7 +158,7 @@ <h3 id="info-cards-title">What's on the ballot</h3>
                         <p class="referendum-description">
                           {{this.description}}
                         </p>
-                      </div>
+                      </div>
```
rogerlaucorp, line 161: Nit: remove this random space. 

1) Random spaces are generally bad style
2) Small misc changes like this end up piling up on bigger PRs, and litter the PR to make it harder to review. 

I'm surprised the linter didn't auto-fix this actually.

src/main/webapp/election-info.js
```diff
@@ -256,4 +256,18 @@ function addPollingStationMarker(map, position, title, description) {
   marker.addListener('click', () => {
     infoWindow.open(map, marker);
   });
-}
\ No newline at end of file
+}
+
+/**
+ * Handlebars helper that removes the beginning "http://" and the trailing "/"
```
rogerlaucorp, line 262: I believe this will also catch "https://". You should comment that as well to be more precise. 

BTW to clarify regarding comments, usually "why" not "what" is the rule, but function docs are a bit of an exception to that, since there's often a lot of subtleties to clarify in functions. Regex is also an exception because nobody can read regex easily lol.

gianelgado12: done

src/main/webapp/electionInfo.html
gianelgado12: done

PR #119, opened 2020-08-17
Add Sources text to bottom of info cards

casprice:


src/main/java/com/google/sps/data/Contest.java
```diff
@@ -84,11 +92,22 @@ public static Contest fromJSONObject(DatastoreService datastore, JSONObject cont
...
     }
 
+    if (contestData.has(SOURCE_JSON_KEYWORD)) {
+      StringBuilder strBuf = new StringBuilder();
+      for (Object sourceObject : contestData.getJSONArray(SOURCE_JSON_KEYWORD)) {
+        JSONObject source = (JSONObject) sourceObject;
+        strBuf.append(source.getString(SOURCE_NAME_JSON_KEYWORD));
+        strBuf.append(", ");
+      }
+      sourceString = strBuf.toString().substring(0, strBuf.length() - 2);
```
julian-yang, line 102: Can you explain why we are doing this?

src/main/java/com/google/sps/data/Referendum.java
```diff
@@ -69,9 +77,20 @@ public static Referendum fromJSONObject(JSONObject obj) throws JSONException {
       referendumDescription = "";
     }
 
+    if (obj.has(SOURCE_JSON_KEYWORD)) {
+      StringBuilder strBuf = new StringBuilder();
+      for (Object sourceObject : obj.getJSONArray(SOURCE_JSON_KEYWORD)) {
```
julian-yang, line 82: This is probably fine.  Though I personally love streams/maps.

I think I would consider using a StringJoiner instead, so that you don't need to have the code below for stripping off the last comma.

https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html

Alternatively, you could also consider storing this as a list and construct the comma separated list on client side.

src/main/java/com/google/sps/data/Contest.java
julian-yang: Oops. ignore this.  But the comment in Referendum.java also applies here

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -41,10 +41,13 @@
...
   public static final String NAME_JSON_KEYWORD = "office";
   public static final String CANDIDATES_JSON_KEYWORD = "candidates";
+  public static final String SOURCE_JSON_KEYWORD = "sources";
+  public static final String SOURCE_NAME_JSON_KEYWORD = "name";
   public static final String NAME_ENTITY_KEYWORD = "name";
   public static final String CANDIDATES_ENTITY_KEYWORD = "candidates";
   public static final String DESCRIPTION_ENTITY_KEYWORD = "description";
-  private static final Logger logger = Logger.getLogger(ServletUtils.class.getName());
+  public static final String SOURCE_ENTITY_KEYWORD = "source";
+  private static final Logger logger = Logger.getLogger(Contest.class.getName());
```
rogerlaucorp, line 50: Nit: place a spce between the private and public variables.

PR #108, opened 2020-08-12
Polling station tests

anooshrees:
- Code now handles passing electionName using servlet calls and dynamic loading with Handlebars instead of storing it in the query params
- Testing for all cases where information is missing from the JSON for a PollingStation object
- Constants for string values in PollingStation and Election
- Handles missing entities in Datastore when calling servlets
- Integration testing for address input page

PR #125, opened 2020-08-18
Test retrieval of entities deleted from Datastore

anooshrees:
Tests the functionality added in PR #108 that handles servlets trying to retrieve Entities in Datastore that have been deleted and no longer exist.


src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -327,9 +331,97 @@ public void oneElection_contestMissingInDatastore_testDoGet() throws IOException
...
+
+  @Test
+  public void oneElection_candidateMissingInDatastore_testDoGet() throws IOException {
+    Entity electionEntity = new Entity("Election");
+    Entity contestEntity = new Entity("Contest");
+
+    Entity firstCandidateEntity = new Entity("Candidate");
+    Entity secondCandidateEntity = new Entity("Candidate");
+
+    electionEntity.setPropertiesFrom(electionEntityOne);
```
julian-yang, line 353: Can you group this with lines 347-348

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -327,9 +331,97 @@ public void oneElection_contestMissingInDatastore_testDoGet() throws IOException
...
+    firstCandidateEntity.setPropertiesFrom(candidateEntityOne);
+    secondCandidateEntity.setPropertiesFrom(candidateEntityTwo);
+
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+    long contestId = ds.put(contestEntity).getId();
+    long firstCandidateId = ds.put(firstCandidateEntity).getId();
+
+    // Delete the Entity from Datastore to mimic what happens at the beginning
+    // of a call to doGet() in ElectionServlet.
+    long secondCandidateId = ds.put(secondCandidateEntity).getId();
```
julian-yang, line 365: I don't quite understand the point of this.  Aren't you just inserting something and then promptly deleting it?

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -327,9 +331,97 @@ public void oneElection_contestMissingInDatastore_testDoGet() throws IOException
...
+    Entity secondReferendumEntity = new Entity("Referendum");
+
+    electionEntity.setPropertiesFrom(electionEntityOne);
+    firstReferendumEntity.setPropertiesFrom(referendumEntityOne);
+    secondReferendumEntity.setPropertiesFrom(referendumEntityTwo);
+
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+    long referendumOneId = ds.put(firstReferendumEntity).getId();
+
+    long referendumTwoId = ds.put(secondReferendumEntity).getId();
```
julian-yang, line 405: Couldn't you just make up some random ID instead?

src/test/java/com/google/sps/servlets/ContestServletTest.java
```diff
@@ -327,9 +331,97 @@ public void oneElection_contestMissingInDatastore_testDoGet() throws IOException
...
+    verify(printWriter)
+        .println(
+            "{\"contests\":[{\"name\":\"myFirstContest\",\"candidates\":[{\"name\":\"myFirstCandidate\","
+                + "\"partyAffiliation\":\"myParty\",\"campaignSite\":\"myWebsite\",\"platformDescription\":"
+                + "\"This is a cool candidate.\",\"channels\":{}}],"
+                + "\"description\":\"This contest is important.\",\"source\":\"Voter Information Project\"}],\"referendums\":[]}");
+  }
+
+  @Test
+  public void oneElection_referendumMissingInDatastore_testDoGet() throws IOException {
```
julian-yang, line 393: can you change this part to describe what the expected result is?
e.g. omits missing referendum or something like that

julian-yang (changes requested):

src/main/java/com/google/sps/data/Contest.java
```diff
@@ -134,7 +134,12 @@ public String toJsonString(DatastoreService datastore) {
                     : JsonNull.INSTANCE)
         // Have to use forEach to have void return. Can't use Collection to collect because
         // JsonArray can't addAll() with String as parameter.
-        .forEach(jsonElement -> candidateJsonArray.add(jsonElement));
+        .forEach(
```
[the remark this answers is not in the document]
rogerlaucorp: +1

src/test/java/com/google/sps/servlets/ContestServletTest.java
anooshrees: This is to imitate what would happen when we do a mass deletion of entities when calling ElectionServlet/InfoCardServlet. Additionally, I can't get a non-zero ID value for the entities (a zero value ID will throw an error when I try to add it to the Contests entity) unless I add to Datastore.

src/test/java/com/google/sps/servlets/ContestServletTest.java
anooshrees: I could, but then I run the risk of generating a random ID that is already assigned to an Entity in Datastore. The deletion imitates what happens in the actual Servlet code.

src/test/java/com/google/sps/servlets/ContestServletTest.java
anooshrees: yes, will do!

anooshrees:
Waiting for confirmation on approach from @julian-yang and then will push commits.

julian-yang (approved):

PR #126, opened 2020-08-19
Referendum URL Functionality

anooshrees:
- Add url to Referendum class and entity storage
- Update Referendum tests
- Add clickable URL to Referendum info cards and format text within to look more substantial


src/main/java/com/google/sps/data/Referendum.java
```diff
@@ -66,6 +72,7 @@ public static Referendum fromJSONObject(JSONObject obj) throws JSONException {
     String referendumDescription;
```
julian-yang, line 72: can you remove the "referendum" prefix?
Or even better, can we just set the properties in the Referendum.builder() directly?

julian-yang (changes requested):

src/main/java/com/google/sps/data/Referendum.java
anooshrees: We use a try/catch when reading the JSON object because the description may or may not exist in the JSON object. Happy to rename the variable, but setting directly in the builder() may cause runtime errors that we can avoid.

julian-yang (approved):

PR #129, opened 2020-08-21
Fill in basic info for the README

casprice:


README.md
```diff
@@ -1 +1,54 @@
...
+```
+mvn package appengine:run
+```
+**Note:** This command must be run in the root directory of the repository, because uses the `pom.xml` file to track dependencies and plugins.
+
+The first time you run this command will take longer to build than usual because Maven is installing the project's dependencies.
+
+### Insert election data in your local Datastore
+```
+curl -X PUT /election
```
julian-yang, line 39: do you need to add "localhost" here?

PR #133, opened 2020-08-25
Key deadlines show on page load and address input is one input field

casprice:
Moved "Key deadlines" window to show when electionInfo.html first loads to slightly speed up buffer time after address is submitted. Also removed all of the extra address input fields to clean up the UI.
Screenshot 2020-08-25 at 11 19 45 AM


casprice:
State name now shows so user knows dates/deadlines are specific to the state they chose.
Screenshot 2020-08-25 at 4 08 08 PM

julian-yang (approved):

PR #143, opened 2020-08-28
Display info text if candidate does not have additional card data

casprice:
image

### DIFF
--- a/src/main/webapp/electionInfo.html
+++ b/src/main/webapp/electionInfo.html
@@ -154,6 +154,12 @@
                               {{/if}}
                               {{#each this.channels}}
                               <p><b>{{@key}}: </b><a href={{this}}>{{stripUrl this}}</a></p>
+                              {{else}}
+                              {{#unless this.campaignSite}}
+                                {{#unless this.platformDescription}}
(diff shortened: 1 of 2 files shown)


src/main/webapp/electionInfo.html
```diff
@@ -154,6 +154,12 @@ <h3 id="info-cards-title">What's on the ballot:</h3>
                               {{/if}}
                               {{#each this.channels}}
                               <p><b>{{@key}}: </b><a href={{this}}>{{stripUrl this}}</a></p>
+                              {{else}}
+                              {{#unless this.campaignSite}}
```
julian-yang, line 158: I'm not really familiar with handlebars, is it possible to combine these two #unless blocks?

julian-yang (approved):

src/main/webapp/electionInfo.html
casprice: I tried looking this up, and found there's no built-in way to do it: https://stackoverflow.com/questions/41251635/handlebars-multiple-conditions-if-statement

But I can make a helper "and" operator which gets the job done the same.

PR #144, opened 2020-08-31
Optimize Datastore Usage

gianelgado12:
To avoid having to delete all non-election entities from datastore with every request for ballot information from the datastore the following have been done:
- Contests and Referendums now have a new field called `division` which specify which division a certain contest/referendum corresponds to.
- Elections now have a new field called `divisions` which is a set of all the electoral divisions queried with respect to the election.
- A query to Google Civic Information API's `representativeInfoByAddress` query is made in `InfoCardServlet` for each ballot information request to determine which electoral divisions the provided address belongs to.
- In `InfoCardServlet`, `voterInfoQuery` is only called when the address provided belongs to one or more districts that are not in the `divisions` field of the current election.
- Using cookies, the information returned by `representativeInfoByAddress` is used in `ContestsServlet` to return only the Contests and Referendums relevant to the user.
- Polling Station Information will no longer be stored on the datastore and instead `PollingStationServlet` now makes a call to `voterInfoQuery` to get the polling station information for every request.


src/main/java/com/google/sps/data/Election.java
```diff
@@ -68,7 +63,8 @@
   // represents their Key names.
   public abstract Set<Long> getReferendums();
 
-  public abstract ImmutableSet<EmbeddedEntity> getPollingStations();
+  // All the divisions that have been queried with this election.
+  public abstract Set<String> getDivisions();
```
rogerlaucorp, line 67: Can this be immutableset?

src/main/java/com/google/sps/data/Election.java
```diff
@@ -229,9 +176,7 @@ public EmbeddedEntity createPollingStation(
    * @return true if contests and Referendums contain elements, false otherwise
```
rogerlaucorp, line 176: update documentation a bit. Should mention divisions now too.

src/main/java/com/google/sps/servlets/InfoCardServlet.java
```diff
@@ -134,8 +162,28 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
...
+        StreamSupport.stream(divisions.spliterator(), false).collect(Collectors.joining("|"));
+    Cookie divisionsCookie = new Cookie("addressDivisions", divisionsString);
+    divisionsCookie.setPath("/contests");
+    divisionsCookie.setVersion(1);
+    divisionsCookie.setMaxAge(SECONDS_PER_HOUR);
+    divisionsCookie.setComment("Electoral divisions the queried address belongs to");
+    return divisionsCookie;
+  }
+
+  /** Returns the realtive complement of secondSet with respect to firstSet. */
```
rogerlaucorp, line 182: typo: relative.

src/main/java/com/google/sps/data/Election.java
rogerlaucorp: Here and below.

src/main/java/com/google/sps/data/Election.java
```diff
@@ -134,92 +130,43 @@ public static Election fromElectionQuery(JSONObject electionQueryData) throws JS
...
       throws JSONException {
     Set<Long> contestKeyList = this.getContests();
     Set<Long> referendumKeyList = this.getReferendums();
-
-    List<EmbeddedEntity> pollingStations = new ArrayList<>(this.getPollingStations());
-    ImmutableSet<EmbeddedEntity> pollingStationSet;
+    Set<String> divisionsList = this.getDivisions();
+    divisionsList.addAll(divisions);
 
     if (voterInfoQueryData.has(CONTESTS_JSON_KEYWORD)) {
```
rogerlaucorp, line 144: This nesting has gotten pretty deep. That's bad, see:
https://drive.google.com/file/d/0B-KFDiTrQe9lLU9uVDZzZWNmRVU/view?usp=sharing

I think we can refactor this like:
------------------

if (!voterInfoQueryData.has(CONTESTS_JSON_KEYWORD)) {
return this; // nothing will get added, so just return early 
}

// ...
for (Object contestObject : contestListData) {
  if (divisions.contains(currentDivision)) {
    continue;
  }

  if (contest.getString(Contest.TYPE_JSON_KEYWORD).equals(Referendum.ENTITY_KIND)) {
    //...
  } else {
    //...
  }
}

gianelgado12: done

src/main/java/com/google/sps/servlets/InfoCardServlet.java
gianelgado12: done

src/main/java/com/google/sps/data/Election.java
gianelgado12: done

src/main/java/com/google/sps/data/Election.java
gianelgado12: done

